PR #27, opened 2020-05-18
DShop monorepo

franckc:
Top level structure is as follow:
 - shop
 - backend
 - packages
 - scripts

Notes:
 - The front-end dependency on @origin/graphql pulls a bunch of unnecessary packages. The plan would be after this PR is merged to update the front-end code to use ether.js and to remove graphql completely.
 - I tested locally and I was able to bring up a store and use the super-admin to create a new shop. But I might have missed things that could break production deployments. So please let me know if you think of anything that could break. 
 - I did not include infra/dshop-control-server infra/dshop-control-client. I was not sure if those are still being used

I'm thinking I should merge this PR once @nick is done with launching the new shops so that we don't risk breaking things and slowing him down.


nick:
Nice 😄 

I just tried a fresh clone, checkout franck/mono, `yarn` and `yarn start` but get:

    UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open '/Users/nick/Projects/dshop/packages/services/../contracts/build/contracts.json'

I guess the services auto-contract deployer isn't working?

mikeshultz:
Just to note: This will impact the VM image generation for marketplace, as well as the heroku deployment, I believe.  

I think for the kubernetes deployment, it will be configured to clone from this repo after merge.

I wonder if `backend` and `shop` should be directories under `packages` to make it more of a typical monorepo?

franckc:
@nick Hmm. thanks for testing! let me test in a fresh repo. I bet you are right and that the marketplace contract doesn't auto-deploy as it should.

@mikeshultz 
- regarding putting everything under packages I don't have a strong opinion. I just copied the DApp structure :) For my information, what's the advantage of having everything under packages? Is that what most mono repos do?
 - yes I'll be careful to monitor the next prod deployment for breakages...
 - do you have a link to the script we currently use for generating the VM image? I can test it...

mikeshultz:
VM image generation stuff is here: https://github.com/OriginProtocol/dshop/tree/franck/mono/backend/scripts/gcp/marketplace

> For my information, what's the advantage of having everything under packages? Is that what most mono repos do?

I think it's convention, but I'm far from an expert in JS packaging.  I think @tomlinton was the one that did the last iteration of the origin monorepo.

tomlinton:
Yep it was just a convention, `packages` was intended to be the packages that we published to npm that other people could build on.

franckc:
Updated a few issues including build contract files missing as reported by @nick 
Also update the VM image generation script to use the new path for the backend.

Regarding moving backend and shop under the packages directory, I'm not opposed to it at all. A couple of observations though:
 - I do like that the current structure makes it clear that the 2 main packages are those two and that everything else under packages are just libraries.
 - Moving shop and backend under packages means pretty much nothing (aside from scripts which has helper shell tools for development) will be left in the root of the repo.
Let me know if anyone feels strongly we should go this direction otherwise I'm inclined to leave as is to start with.

backend/scripts/gcp/marketplace/startup_script.sh
```diff
@@ -0,0 +1,182 @@
...
+chown www-data:www-data /etc/resty-auto-ssl
+openssl req -new -newkey rsa:2048 -days 3650 -nodes -x509 \
+    -subj '/CN=sni-support-required-for-valid-ssl' \
+    -keyout /etc/ssl/resty-auto-ssl-fallback.key \
+    -out /etc/ssl/resty-auto-ssl-fallback.crt
+
+###############
+# Install dshop
+###############
+BACKEND_PATH="/app/dshop/backend"
```
mikeshultz, line 37: Lacks `export` this will be undefined when used later.  Perhaps skip using vars here since it's only used once.

mikeshultz:
I don't feel very strongly about it either, but I'll play devil's advocate:

> I do like that the current structure makes it clear that the 2 main packages are those two and that everything else under packages are just libraries.

They don't look like packages to me.

I tried to review the VM image startup script but this is an exceptionally difficult PR to review.  Let me know if there's anything in particular I should be looking at.

packages/contracts/build/contracts.json
```diff
@@ -0,0 +1,22 @@
...
+    "Admin": "0x627306090abaB3A6e1400e9345bC60c78a8BEf57",
+    "Seller": "0xf17f52151EbEF6C7334FAD080c5704D77216b732",
+    "Buyer": "0xC5fdf4076b8F3A5357c5E395ab970B5B54098Fef",
+    "Arbitrator": "0x821aEa9a577a9b44299B9c15c88cf3087F3b5544",
+    "Affiliate": "0x0d1d4e623D10F9FBA5Db95830F7d3839406C6AF2",
+    "OGN": "0xF12b5dd4EAD5F743C6BaA640B0216200e89B60Da",
+    "DAI": "0x345cA3e014Aaf5dcA488057592ee47305D9B3e10",
+    "OKB": "0xf25186B5081Ff5cE73482AD761DB0eB0d25abfBF",
+    "USDT": "0x8f0483125FCb9aaAEFA9209D8E9d7b9C8B9Fb90F",
+    "Marketplace": "0xFB88dE099e13c3ED21F80a7a1E49f8CAEcF10df6",
```
nick, line 11: This file should be generated locally rather than checked in, as all the addresses refer to contracts on the local testnet

backend/scripts/gcp/marketplace/startup_script.sh
franckc: Ah yes. Thank you, great catch!

packages/contracts/build/contracts.json
franckc: I'm following the same pattern as in the DApp where those files are checked in but overwritten at startup. See https://github.com/OriginProtocol/origin/tree/master/packages/contracts/build

The challenge with removing those files is that the [contract deployer](https://github.com/OriginProtocol/origin/blob/master/packages/services/index.js#L184) is a graphql server and it can't get started without those files as otherwise it complains about missing imports! See this [line](https://github.com/OriginProtocol/origin/blob/master/packages/graphql/fixtures/populate-server.js#L15).

I'd suggest that for the short-term we go with this but as a separate PR it would be good to simplify contract deployment (especially since it is only 1 contract, the marketplace_v1) so that it does not depend on graphql anymore but just runs truffle.

backend/scripts/gcp/marketplace/startup_script.sh
franckc: Actually, you only need to export for child processes to get the env var.
In this case the var is only used in the script, so unless I'm missing something there is no need to export?

mikeshultz: :shrug: it was really early.  Still looks unnecessary to me.

PR #32, opened 2020-05-20
Record dshop deployments in the DB

franckc:
This was originally a PR on the origin repo: <https://github.com/OriginProtocol/origin/pull/4468>

For every shop created, the following info is recorded by the back-end in a new "shop_deployments" table:
 - IPFS hash of the store
 - domain
 - IPFS gateway

I tested on my local and the fields were populated.


backend/models/shop_deployment.js
```diff
@@ -0,0 +1,21 @@
+module.exports = (sequelize, DataTypes) => {
+  const ShopDeployment = sequelize.define(
+    'ShopDeployment',
+    {
+      shopId: DataTypes.INTEGER,
+      domain: DataTypes.STRING,
+      ipfsGateway: DataTypes.STRING,
+      ipfsHash: DataTypes.STRING
```
mikeshultz, line 8: Might want to add created/updated so they can be used?

franckc: I don't think you need to declare those explicitly in the model. Sequelize does it for you automagically :) https://sequelize.org/v5/manual/models-definition.html#timestamps

franckc:
Merging this in master. Happy to address any further comments as a subsequent PR.

PR #36, opened 2020-05-22
Rename “stripeKey” to “stripePublishableKey”

DanielVF:
(Copied PR over from dapp monorepo.)

The field name "StripeKey" is could mean either the the secret key or the public key. I’m concerned with the old name that someone will put in their secret key into this field and cause a mess.

I’ve given this a gentle fallback - the old key is still used, and a warning shown in the /admin/settings/client tab.


DanielVF:
Since this is old, just closing.

PR #37, opened 2020-05-22
Remove GraphQL

nick:


packages/contracts/package.json
```diff
@@ -20,27 +20,40 @@
   },
   "license": "MIT",
   "dependencies": {
```
franckc, line 22: For my information, why do we need these dependencies (rather than devDependencies) given that aside from the static JSON config files with contract addresses for Mainnet/Rinkeby, nothing else from the contract package should be needed for a production build?

nick: Can probably move them to devDependencies, yea

shop/src/data/getOffer.js
```diff
@@ -1,49 +1,115 @@
...
+    ],
+    payable: false,
+    stateMutability: 'view',
+    type: 'function'
+  }
+]
+window.provider = provider
+const marketplaceInterface = new ethers.utils.Interface(marketplaceAbi)
+
+const marketplaceAddr = '0x9FBDa871d559710256a2502A2517b794B482Db40'
```
franckc, line 67: Should we get this from the network config rather than hardcoding it?

packages/contracts/package.json
nick: I think this was a bad merge that caused these to move back as I was updating package versions

PR #57, opened 2020-05-29
[Backend] Use JSONB rather than TEXT for DB cols storing JSON data

franckc:
Migrate 3 tables to using JSONB rather than TEXT for columns storing JSON data:
 - orders
 - external_orders
 - external_payments

Added a unit test (our very first backend unit test!) to test the changes. It can be run using ```yarn run test```

Notes regarding sqlite:
 - JSONB is not supported so we use JSON
 - altering a column type is messy (involves creating a temp table). since sqlite is not used in prod, we create the column with the JSON type in the first place and skip the migration.


backend/models/external_orders.js
```diff
@@ -11,7 +11,7 @@ module.exports = (sequelize, DataTypes) => {
       updated_at: DataTypes.DATE,
       external_id: DataTypes.STRING,
       order_id: DataTypes.STRING,
-      data: DataTypes.TEXT,
+      data: DataTypes.JSON, // Note: unlike JSONB, JSON is supported by both Postgres and sqlite.
```
mikeshultz, line 14: So, for sqlite, the model type is defined as `JSON` but the migration type(actual table type) is `TEXT`?  This works?

franckc: Oops. Updated this model - it was an older version. Good catch.

Here are the 2 scenarios:
 - For Postgres: column was created as TEXT originally, migrate to JSONB. Model uses JSONB.
 - For sqlite: column gets created as JSON, migration gets skipped. Model uses JSON.

Lmk if that does not make sense...

mikeshultz: I guess that makes sense?  What about any existing sqlite installations(e.g. heroku or external users)?

franckc: Good question. Sqlite was added only recently a couple of weeks ago (as part of the work Daniel did for creating a standalone VM image). The VM images is not used externally afaik. Regarding Heroku the default button deployment used Postgres so I think we should be safe as well.

mikeshultz:
SQLite support is much older than that

backend/models/external_orders.js
franckc: Oh wow. My bad. @mikeshultz you are totally right... I did some archeology and sqlite has been there actually forever. It was in the very first commit for dshop :) https://github.com/OriginProtocol/origin/pull/4022

Anyway, I'm not aware of anyone else than us running any dshop? ...So I'm not too worried about this non-backward compatible change for sqlite, but if anyone has any objections, please let me know.

PR #61, opened 2020-06-02
[Listener] Filter out events in logic

franckc:
The listener now subscribes to all events emitted by the marketplace contract and filters out events that are not related to dshop in the logic.

Previously the listener was filtering out events by specifying the dshop listing ids in the query it was making to the network. This resulted in the possible issue:
 1. Shop creation flow start from superadmin, a listing is created on the marketplace
 2. Listener picks up the new listing event but the listing is not associated yet with the new shop in the DB (that only happens in step 3 below). So the listener ignores the event.
 3. Shop creation flow finishes, listing id gets associated with shop in the DB
 4. Transactions for the new shop are ignored since the listener is not listening for events on the new shop's listing

Added unit tests to check the new logic.


backend/scripts/backfill.js
```diff
@@ -181,7 +181,7 @@ async function handleEvents(listingIdFull) {
     order: [['block_number', 'ASC']]
   })
   for (const event of events) {
-    await insertOrderFromEvent({
+    await processDShopEvent({
       offerId: `${listingIdFull}-${event.offerId}`,
```
nick, line 185: I guess we don't need this any more since it may not be an offer event?

nick:
Could it also be worth adding a test for a non Order event, like CreateListing? Just to make sure that doesn't mess with anything.

backend/scripts/backfill.js
franckc: Good point. Removed the offerId argument from this method.

franckc:
Addressed review comments, added a test for non-Offer event ListingUpdated.

PR #73, opened 2020-06-08
Adds support for ipfs-cluster to shop deployer

mikeshultz:
Adds support for ipfs-cluster to shop deployer.  It currently uses the same host as the network-level configured IPFS API, you just need to add the user and password.


mikeshultz:
@nick So dealing with merge conflicts I noticed the new `pinner` argument to `deployShop()`.  I see it's tied to a select box.  I wonder if it's useful to select the pinner service here, rather than a boolean "should be pinned" kind of setting?

If pinata and an IPFS cluster is configured, why not pin to both when doing deployment?

nick:
Sure we can merge this and I'll convert the select to checkboxes instead

mikeshultz:
Okay.  Added `ipfs-cluster` option to the dropdown anyway to match the current implementation.

PR #84, opened 2020-06-11
Store user creation was not refreshing users list

DanielVF:
Store user creation was not refreshing sellers list. Added code to refresh after creation. Cleaned up seller creation code.


backend/routes/auth.js
```diff
@@ -173,9 +173,17 @@ module.exports = function (app) {
 
   app.post('/auth/logout', logoutHandler)
 
-  // TODO: Should this at least use API key auth?
   app.post('/auth/registration', async (req, res) => {
-    const { seller, status, error } = await createSeller(req.body)
+    if ((await numSellers()) > 0) {
+      return res.status(status).json({
```
franckc, line 178: status is not set at that point. and anyway it should just return a 200 which I think it does by default so perhaps just change this to 
`return res.json({...})`

backend/utils/sellers.js
```diff
@@ -1,10 +1,11 @@
 const { Seller } = require('../models')
 const { createSalt, hashPassword, checkPassword } = require('../routes/_auth')
 
-async function createSeller({ name, email, password, superuser }) {
+async function createSeller({ name, email, password }, opts) {
   if (!name || !email || !password) {
     return { status: 400, error: 'Invalid registration' }
```
franckc, line 6: Not modified in this PR, but should this be
```
return res.status(400).json({ error: 'Invalid registration' })
```

Same comment on line 14 below.

backend/utils/sellers.js
```diff
@@ -14,20 +15,23 @@ async function createSeller({ name, email, password, superuser }) {
     return { status: 409, error: 'Registration exists' }
   }
 
-  const numSellers = await Seller.count()
   const salt = await createSalt()
   const passwordHash = await hashPassword(salt, password)
 
   const seller = await Seller.create({
     name,
     email,
```
franckc, line 23: Should email get lowercased when storing it since line 11 above relies on that?

backend/utils/sellers.js
```diff
@@ -14,20 +15,23 @@ async function createSeller({ name, email, password, superuser }) {
...
 
   return { seller }
 }
 
+async function numSellers() {
+  return await Seller.count()
+}
+
 async function findSeller(email) {
   const seller = await Seller.findOne({ where: { email } })
```
franckc, line 36: Document that the email arg should be lower case or have this method lower case when doing the query.

backend/utils/sellers.js
DanielVF: Now lowercasing all emails, both for reading and for writing. Postel's law.

backend/utils/sellers.js
DanielVF: Lowercased.

backend/utils/sellers.js
DanielVF: This function is not a route in itself - it's more of a model layer thing. Since it doesn't return to the view, I think it should stay the same as it is now.

backend/routes/auth.js
DanielVF: Done. Changed to a return a 409 to be consistent with other dshop code's behavior when trying to create a duplicate.

PR #90, opened 2020-06-13
Nick/ui changes

nick:


nick:
* Merged master
* Lint
* Removed react-bootstrap dep
* A bunch of styling updates
* Used new ConfirmationModal component
* Replaced data/api with new useBackendApi hook

shop/src/components/ConfirmationModal.js
```diff
@@ -0,0 +1,98 @@
...
+        <Modal
+          shouldClose={state.shouldClose}
+          onClose={() => {
+            if (state.confirmed) {
+              onSuccess()
+            } else {
+              setState({}, true)
+            }
+          }}
+        >
```
shahthepro, line 48: Any thoughts on using Bootstrap model? I know the plan was to switch to them on DApp. But, was wondering if it is also on Dshop pipeline. 

I did add `react-boostrap` package and used the bootstrap Modal in a few places. Can revert that back if we are going to stick with this custom written component

nick: The modal in components/Modal is re-written to use the Bootstrap styles, so is an improvement on the one in the DApp. Wanted to avoid react-bootstrap as I think it's a fairly big dependency

shahthepro: Got it. Didn't think about the size of react-bootstrap. Will remove it then

Issue #12, opened 2020-05-08
Deploy Veil merch store


gullible1:
All Veil merch has been [created in Printful](https://www.printful.com/dashboard/sync?store=4511689), and we need to deploy the Veil store in order to share it with their team and get a final sign off prior to launch.

PR #108, opened 2020-06-19
Logger

mikeshultz:
Use logplease instead of console logging


backend/utils/logger.js
```diff
@@ -1,16 +1,29 @@
...
+    showTimestamp: !isProd,
+    useColors: !isProd
+  })
+}
 
-const listenerLogger = Logger.create('listener', {
-  showTimestamp: false
-})
+const rootLogger = getLogger()
+const apiLogger = getLogger('dshop.api')
```
franckc, line 21: given line #12, isn't that going to create logger dshop.dshop.api?

mikeshultz: yeah, and actually, I think these can just be removed.  There's nothing else to keep compatibility with.  Thanks

PR #113, opened 2020-06-22
Listener stuff

mikeshultz:
This PR is just a placeholder for now.  Want to see what Alchemy comes back with for eth_getLogs block range params before further work is done on this.


mikeshultz:
Trashing this PR.  Doesn't look like we need to make any changes.

Issue #118, opened 2020-06-23
IFPS issuer cert priming


mikeshultz:
Look into "priming" the IPFS issuer on new domains.  By doing this, we can have our backend trigger an ACME cert issuance before the user tries it saving them a minute or two of extra load time on their first store load.  This might even be doable before the IPFS deploy is complete.  Need to see how the autossl issuer acts when the request is a 404.

There's potentially some gotchas here.  The DNS changs must propagate to Let's Encrypt's DNS servers(and not be cached). If the issuer makes the ACME request early and LE(or their upstream) isn't updated, it could end up caching the old data for up to almost double the TTL(TTL of the new request landing the end of the TTL of the original old data).

Worth at least testing to see.  I'm not even sure LE's internal DNS does any TTL caching, just making assumptions.  In my past use, it's been pretty quick to notice changes.

Ref #115

mikeshultz:
autossl still works when it's a 404. This can be done ass soon as DNS is configured and resolving.

PR #131, opened 2020-06-24
AutoSSL priming

mikeshultz:
This will make a request to the issuer to try and trigger AutoSSL before the user does.  It's a bit of a race between this and the user with the way things are right now.  It takes a couple of minutes for Let's Encrypt to resolve the new DNS record.

This also changes the DNS configuration to use the network-level configured IPFS gateway instead of hardcoded values.  We need to make sure that's pointed at the issuer(it is on primary deployments, prod and staging).

There's a reasonable chance this will do the thing before the user, but think we might still want to add some UI stall tactics(like a spinner!).

Ref: #118

### DIFF
--- a/backend/utils/deployShop.js
+++ b/backend/utils/deployShop.js
@@ -298,6 +298,7 @@ async function deployShop({
     domain = dnsProvider ? `https://${subdomain}.${zone}` : null
     await configureShopDNS({ network, subdomain, zone, hash, dnsProvider })
     if (network.ipfs) {
+      // Intentionally not awating on this so we can return to the user faster
       triggerAutoSSL(domain, network.ipfs)
     }
   }
--- a/backend/utils/deployShop.js
+++ b/backend/utils/deployShop.js
@@ -65,7 +65,7 @@ async function triggerAutoSSL(url, autoSSLHost) {
 
   log.debug(`Making request to ${url.hostname} with header (Host: ${hostname})`)
 
-  while (!success && attempts <= AUTOSSL_MAX_ATTEMPTS) {
+  for (;;) {
     try {
       await getHTTPS({
         host: url.hostname,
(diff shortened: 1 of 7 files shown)


backend/utils/deployShop.js
```diff
@@ -44,6 +47,70 @@ function urlToMultiaddr(v, opts) {
...
+      )
+
+      if (attempts <= AUTOSSL_MAX_ATTEMPTS) {
+        // sleep with backoff
+        await (async () => {
+          log.debug(
+            `AutoSSL trigger backing off by ${attempts * AUTOSSL_BACKOFF}ms`
+          )
+          return new Promise((resolve) =>
+            setTimeout(resolve, attempts * AUTOSSL_BACKOFF)
```
franckc, line 94: Worst case scenario, that's a wait of 30s + 2 * 30s + 3 * 30s + 4 * 30s + 5 * 30s = 7.5min which seems a bit long?
Should AUTOSSL_BACKOFF get lowered a bit?

backend/utils/deployShop.js
```diff
@@ -228,6 +297,9 @@ async function deployShop({
   if (subdomain) {
     domain = dnsProvider ? `https://${subdomain}.${zone}` : null
     await configureShopDNS({ network, subdomain, zone, hash, dnsProvider })
+    if (network.ipfs) {
+      triggerAutoSSL(domain, network.ipfs)
```
franckc, line 301: I assume we intentionally want to run this in the background so no await.
Perhaps add a comment explaining this.

backend/utils/deployShop.js
mikeshultz: The backoff was chosen specifically because it was > 5min.  This is intentional.

backend/utils/deployShop.js
mikeshultz: Will add a note

PR #138, opened 2020-06-26
Add DNS verification endpoint

shahthepro:


backend/routes/domains.js
```diff
@@ -0,0 +1,68 @@
...
+      log.debug('Trying to verify DNS records of domain', domain, hostname)
+
+      try {
+        const cnameData = await new Promise((resolve, reject) => {
+          resolver.resolveCname(domain, (err, data) =>
+            err ? reject(err) : resolve(data)
+          )
+        })
+
+        const cnameValid = cnameData.includes('ipfs-prod.ogn.app')
```
mikeshultz, line 29: We need to stop using this hardcoded value.  How about using `network.ipfs` from the network table, since that's what's being used in the DNS records assignments now?

shahthepro: Thanks @mikeshultz. Will make these changes

PR #184, opened 2020-07-09
Log printful events to DB and sync on product_updated event

shahthepro:


backend/db/migrations/20200709112529-addExternalEvents.js
```diff
@@ -0,0 +1,28 @@
...
+      id: {
+        type: Sequelize.INTEGER,
+        primaryKey: true,
+        autoIncrement: true
+      },
+      shop_id: {
+        type: Sequelize.INTEGER,
+        references: { model: 'shops', key: 'id' }
+      },
+      service: Sequelize.STRING,
```
franckc, line 17: I think it would be nice to use enums rather than strings.
It makes it easier to understand the schema.
Something we discussed in Tom's productivity doc and earlier this week during the offsite as part of the "best coding practices" we would like to apply moving forward.

backend/routes/shops.js
```diff
@@ -413,10 +411,15 @@ module.exports = function (app) {
     log.info(`Outputting to ${OutputDir}`)
 
     if (shopType === 'printful' && printfulApi) {
-      await downloadProductData({ OutputDir, printfulApi })
-      await writeProductData({ OutputDir })
-      await downloadPrintfulMockups({ OutputDir })
-      await resizePrintfulMockups({ OutputDir })
+      // Should this be made async? Like just moving to the queue instead of blocking?
```
franckc, line 414: Good point. How long do these operations take. I'd assume for a shop with a large number of products it could take a while?

shahthepro: @franckc We fetch just first 100 products (and then write the data for each of the fetched products to separate files). At most, it can be 1 API calls and 100 file write. But we also download, resize and move around the image mockups for each of the products. I'm not sure what's the max number of images/mockups per product though. So, could take longer if you have a really large list of products  (each with a significant number of variants and mockups).

backend/db/migrations/20200709112529-addExternalEvents.js
shahthepro: Will change it, thanks

backend/routes/shops.js
franckc: Ah... good to know there is a 100 products limit. thank you for explaining.

Perhaps you could add some log.info at the start and end of the operation statement so that we can see how long it take by looking at the logs and later move to a queue if we really it takes a while to process?

shahthepro: That sounds good. Will log the time taken for each of these ops.

shahthepro: Sorry, I was wrong about 1 API call. We actually make a couple more API calls for each of the product. So, it will 201 API at most.

PR #193, opened 2020-07-09
Multi instsance listener support

mikeshultz:
Basically just throws `handleLog` into a bull queue processor.  Could integrate it further in the future(e.g. job progress logging) but this will get things working.

Works pretty alright on my local.


backend/listener.js
```diff
@@ -129,22 +129,32 @@ async function connectWS({ network }) {
...
+        await eventsQueue.add(
+          {
+            ...result,
+            address,
+            networkId,
+            contractVersion
+          },
+          {
+            attempts: 3,
+            jobId: result.transactionHash
```
franckc, line 141: nit: it's not obvious for folks not deeply familiar with bull so would be useful to have a comment explaining that the bull queue is idempotent and that as multiple instances of the listener enqueue the same jobId, only one gets executed.

backend/listener.js
```diff
@@ -167,7 +177,17 @@ const handleNewHead = (head, networkId) => {
   const number = web3.utils.hexToNumber(head.number)
   const timestamp = web3.utils.hexToNumber(head.timestamp)
 
-  Network.upsert({ networkId, lastBlock: number })
+  Network.update(
+    { networkId, lastBlock: number },
```
franckc, line 181: you did not change it but I wonder why is it necessary to update the networkId?

backend/listener.js
```diff
@@ -167,7 +177,17 @@ const handleNewHead = (head, networkId) => {
...
+    {
+      where: {
```
franckc, line 183: Oh cool, I did not realize an upsert can have a where clause. 

Do you know if it is atomic? In other words, does the selection and the update of the row happens at once?

Otherwise there is a chance of a race condition which could cause the cursor to backtrack.
For example:
 - lastBlock is at  10
 - instance A calls upsert to update it to 20
 - concurrently instance B calls upsert to update it to 15
=> depending on timing cursor may get updated to 15 or 20.

Probably not a big issue since it may just cause some logs to get picked up again (for example if the listener gets restarted) but with the idempotent properties of the queue it should not cause a log to be processed more than once.

Another way to do it that is atomic would be to use a FOR UPDATE statement to lock the row, then check if lastBlock < number, then update the row.

backend/listener.js
mikeshultz: Sequelize will insert a row with a null `network_id` if it's not included...

backend/listener.js
mikeshultz: Your suggestion that a race condition can happen to cause a lower block number to be saved to the curosor should not be possible due to the added WHERE clause that you're commenting on.  DML statements should all be evaluated sequentially.  

I didn't ignore your comment originally suggesting the "FOR UPDATE" statement.  In fact, I burned a few hours trying to implement it with Sequelize just to realize that SQLite does not support row level locking and doing this causes a full DB lock in SQLite causing issues across the board.  I believe the worst case scenario that can happen with my change is two job inserts into bull, which has no effect.

If you want to iterate on this later and tighten it up feel free.

backend/listener.js
```diff
@@ -129,22 +129,32 @@ async function connectWS({ network }) {
...
-          web3,
+        await eventsQueue.add(
+          {
+            ...result,
+            address,
+            networkId,
+            contractVersion
+          },
+          {
+            attempts: 3,
```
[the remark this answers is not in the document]
mikeshultz: Small chance of an SQL error if it attempts to insert an order that already exists if the printful fulfill throws but that just means it'll error and stop every execution.  I'm not seeing anything serious here.

backend/listener.js
mikeshultz: Sure, added a comment.

backend/listener.js
franckc: Ah thanks for considering it. I did not know about SQLlite not supporting FOR UPDATE, good to know.

I don't think this is true "DML statements should all be evaluated sequentially." given that Postgres is multi-threaded and multiple queries run concurrently?

But agreed it's not a big deal. Fine to leave as is.
I'm a bit OCD about those locking things because it burnt me many times when working on other projects that had high concurrency :)

franckc:
LGTM. Let's get this in?

backend/listener.js
mikeshultz: I think that would violate ACID compliance for DBs?  I can't imagine both WHEREs being evaluated at the same time and then the actual data write happening causing that race.  That would be mayhem!  

But I had some doubts now that you brought it up so I went down the rabbit hole and [found this](https://tapoueh.org/blog/2018/06/postgresql-concurrency-data-modification-language/):

> As row locking is done per-tuple in PostgreSQL, an update statement only ever blocks another update, delete or select for update statement that targets the same row(s).

This isn't official docs or anything, but I [also found this](https://www.postgresql.org/message-id/CAKFQuwaH2o4Qrbr42qsc_AjsM99p_TDvuNfhLimEUt_jhJS%3Dfg%40mail.gmail.com) which suggests there are implicit locks that happen during an update. He also suggests that would be including the evaluation of the WHERE clause since he explicitly mentions doing the evaluation separately via SELECTs would cause a race like you mention.  I'm like 98% confident that these UPDATEs with the WHERE clause checking the updated value should prevent that scenario you presented.

I'm not really sure about the behavior of SQLite on the other hand(supposedly it's ACID compliant), but that DB is unlikely to ever be used by a multi-instance setup except for local testing. Hopefully.

Issue #168, opened 2020-07-06
Re-deploying an unmodified store causes a crash in the backend


franckc:
If a store does not have any changes, its hash remains the same and any attempt to redeploy it causes the following error on the back-end:
```error: duplicate key value violates unique constraint "shop_deployment_names__ipfs_hash__hostname__idx"```
This is due to the following insert sql statement:
```INSERT INTO "shop_deployment_names" ("id","ipfs_hash","hostname","created_at","updated_at") VALUES (DEFAULT,$1,$2,$3,$4) RETURNING *;```

For example:
```
2020-07-06T21:03:54.776019382Z       'INSERT INTO "shop_deployment_names" ("id","ipfs_hash","hostname","created_at","updated_at") VALUES (DEFAULT,$1,$2,$3,$4) RETURNING *;',
2020-07-06T21:03:54.776027254Z      parameters:
2020-07-06T21:03:54.776033247Z       [ 'QmWC18pdphU9Xe9gDES238m4WHdKp9BEpEZM7by6rANVdq',
2020-07-06T21:03:54.776040311Z         'francks-shop.shopwithslice.com',
2020-07-06T21:03:54.776047079Z         '2020-07-06 21:03:54.763 +00:00',
2020-07-06T21:03:54.776053764Z         '2020-07-06 21:03:54.763 +00:00' ] 
```

I think there could be some legitimate reasons to re-deploy a store. For example re-priming IPFS or deploying using a different IPFS cluster, or wanting to change the DNS for a store.

So it seems it would be best if we would support this scenario gracefully in the backend by just updating shop_deployment_names rather than trying to insert into it.

mikeshultz:
Not a crash

franckc:
Fixed. Closing.

PR #196, opened 2020-07-10
Deployment fix and cleanup

franckc:
Fix for #168 and some cleanup of the deployment code.


backend/db/migrations/20200710060151-alterShopDeployment.js
```diff
@@ -0,0 +1,16 @@
...
+  up: queryInterface => {
+    return queryInterface.sequelize.query(`
+      ALTER TABLE shop_deployments RENAME COLUMN ipfs_gateway TO ipfs_pinner;
+    `)
+  },
+  down: queryInterface => {
+    if (queryInterface.sequelize.options.dialect === 'sqlite') return Promise.resolve()
+
+    return queryInterface.sequelize.query(`
+      ALTER TABLE shop_deployments RENAME COLUMN ipfs_pinner TO ipfs_gateway;
```
mikeshultz, line 13: Is this accurate?  All of the entries in master and staging are, in fact, a gateway URL: Specifically: `https://gateway.pinata.cloud`

backend/utils/deployShop.js
```diff
@@ -223,62 +237,72 @@ async function deployShop({
       .replace('FAVICON', publicShopConfig.favicon || 'favicon.ico')
   )
 
-  const remotePinners = []
+  // Note: for legacy reasons, the URLs for the IPFS Gateway and API are stored in
```
mikeshultz, line 240: Is it legacy reasons?  Seems like something that doesn't really require encryption.

backend/db/migrations/20200710060151-alterShopDeployment.js
franckc: Yes currently the code populates the gateway URL. After this change is deployed to prod I'm planning on updating the legacy entries to point to the pinner URL so that the data is consistent.

backend/utils/deployShop.js
franckc: Good point. Hmmm... I don't think we are very clear about what should go as a column vs as encrypted in the config blob :( That's why I said "legacy reasons" :)

It seems we are not consistent: the config right now includes a mix of data that does / does not require encryption. For example things like "dataUrl", "publicUrl", "bigQueryTable" are stored in the config but are not security sensitive...

PR #197, opened 2020-07-10
Sync only updated products

shahthepro:


backend/db/migrations/20200710134411-changeEventTypeToEnum.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
+const { PrintfulWebhookEvents, ExternalServices } = require('../../utils/enums')
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
```
franckc, line 6: ah cool. appreciate you updating this to an enum. will make our schema cleaner :)

backend/utils/enums.js
```diff
@@ -0,0 +1,19 @@
+const PrintfulWebhookEvents = {
```
franckc, line 1: nit: do you mind adding a comment to remind developers that if they make any change to those definitions, they also need to add a DB migration file to add/delete whatever enum value they are adding/deleting?

backend/scripts/printful/downloadProductData.js
```diff
@@ -1,20 +1,61 @@
...
+
 const getProducts = require('./getProducts')
 const getProduct = require('./getProduct')
 const getProductIds = require('./getProductIds')
 
 const PrintfulURL = 'https://api.printful.com'
 
-async function downloadProductData({ OutputDir, printfulApi }) {
+/**
+ * Download's product data from Printful
```
franckc, line 12: thanks for adding those jsdoc comments! super helpful.

Super nit: type. Downloads vs Download's

PR #213, opened 2020-07-14
Sentry back-end integration

franckc:
Integrate with Sentry to capture:
 - API server errors
 - Listener errors

Refactored the API server to use https://github.com/express-promise-router.
This now sends uncaught exceptions thrown within any route to the custom error handler and also properly sends a 500 error back. Before this change, the exceptions were not caught (Express 4 does not do it automatically) and that would cause the server to not return any response (e.g. the front-end would hang waiting for a response ).

### DIFF
--- a/backend/sentry.js
+++ b/backend/sentry.js
@@ -1,0 +1,13 @@
+const Sentry = require('@sentry/node')
+const sentryEventPrefix =
+  'https://sentry.io/organizations/origin-protocol/projects/squad-deals-api/events'
+
+Sentry.init({
+  dsn: process.env.SENTRY_DSN,
+  environment: process.env.ENVIRONMENT || 'development'
+})
+
+module.exports = {
+  Sentry,
+  sentryEventPrefix
+}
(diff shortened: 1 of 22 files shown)


backend/sentry.js
```diff
@@ -0,0 +1,25 @@
...
+const log = getLogger('sentry')
+
+const dsn = process.env.SENTRY_DSN
+const environment = process.env.ENVIRONMENT || 'development'
+
+if (dsn) {
+  log.info(`Initializing Sentry with environment: ${environment}`)
+  Sentry.init({ dsn, environment })
+} else {
+  if (environment === 'development') {
```
mikeshultz, line 15: Mentioned this in the other PR late, so not sure if you saw it, but why have this check at all?  Why not init sentry if `SENTRY_DSN` is defined?  Don't think we gain anything but an extra var to configure from this check.

franckc: What I wanted to guard against is SENTRY_DSN not being set in production and us no noticing it...

Regarding the ENVIRONMENT env var, we do need it anyway to distinguish in Sentry between errors logged from staging vs prod.

mikeshultz: Read through this fast and only looked at the log statement that says "Skipping."  Message is a little misleading but the logic is all good.  :+1:

franckc:
Merging this in to test on staging...

PR #244, opened 2020-07-16
[offer processor] Default to using the network PK if the shop's config does not have a PK

franckc:
- Default to using Network PK
 - Capture exceptions in Sentry

### DIFF
--- a/backend/queues/makeOfferProcessor.js
+++ b/backend/queues/makeOfferProcessor.js
@@ -32,9 +32,10 @@ async function processor(job) {
     job.progress(progress)
   }
 
+  const { shopId, encryptedData, paymentCode } = job.data
+  log.info(`Creating offer for shop ${shopId}`)
+
   try {
-    const { shopId, encryptedData, paymentCode } = job.data
-    log.info(`Creating offer for shop ${shopId}`)
     const shop = await getShop(shopId)
     queueLog(5, 'Load encrypted shop config')
     const network = await getNetwork(shop.networkId)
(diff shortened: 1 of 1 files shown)


nick:
Hrm not sure I see the relevant change in here... looks like we just added sentry?

backend/queues/makeOfferProcessor.js
```diff
@@ -23,46 +27,55 @@ function attachToQueue() {
...
+    queueLog(10, 'Creating offer')
+    const lid = ListingID.fromFQLID(shop.listingId)
+    const offer = createOfferJson(lid, encryptedData, paymentCode)
+    const ires = await postOfferIPFS(network, offer)
+
+    queueLog(20, 'Submitting Offer')
+    const web3 = new Web3(network.provider)
+
+    // Use the Shop PK if there is one, otherwise fall back to Network PK.
+    const backendPk = shopConfig.web3Pk || networkConfig.web3Pk
```
mikeshultz, line 54: Maybe we should make this more defensive?  What if a deployment doesn't have a privkey configured for the network yet?

franckc: good to be defensive. let me check and throw an error...

franckc: added

mikeshultz: That doesn't address my concern above?  What happens if neither `shopConfig.web3Pk` nor `networkConfig.web3Pk` are anything on someone's custom deployment?

PR #247, opened 2020-07-16
AutoSSL "Priming" moved to queue

### LINKED ISSUE #164: AutoSSL priming is getting GCed
It appears that the AutoSSL priming functionality is getting garbage collected and stops processing.  A recent test on staging showed that only one attempt was made before deployment was done.  This differs from local testing because uploading and pinning is faster from our deployed backends which gives less time for this priming to complete.

mikeshultz:
Fixes #164


backend/queues/autosslProcessor.js
```diff
@@ -0,0 +1,33 @@
...
+    job.progress(progress)
+  }
+
+  log(10, 'Making request AutoSSL priming request...')
+
+  const success = await triggerAutoSSL(job.data.url, job.data.host)
+
+  if (!success) {
+    const errMsg = 'AutoSSL priming failed'
+    logger.error(errMsg)
```
franckc, line 26: I wonder if there is a way with Bull to tell if it is the final retry and in that case to capture the exception with Sentry.captureException() so that we get a sense of how often it fails.

mikeshultz: Well, I think we can get that info from bull, so we won't have to give it to it. However, I think maybe every one should be added to Sentry, not just the final one(analytics, not notifications).  

But no, I didn't consider adding Sentry with this.  Just refactoring existing code.  Added

mikeshultz: Actually, I don't like the idea of sending every failed job to sentry because, at least with this queue, it's almost guaranteed to error at least the first couple attempts.  I added a `failed` error handler on all of the queues to add error capture.  This might impact your other sentry efforts(possible duplication).  Thoughts?

franckc: ah yeah your way of doing it is more generic. right now some exceptions that I log will be duplicated.
https://github.com/OriginProtocol/dshop/blob/master/backend/queues/eventsProcessor.js#L30
https://github.com/OriginProtocol/dshop/blob/master/backend/queues/makeOfferProcessor.js#L83

no big deal - I can clean up

mikeshultz: Okay, I added a bit of configuration to enable this failed capture.  For those examples you showed, if they were converted to use the event handler, it seems like the `error` might be more appropriate than `failed`.

PR #248, opened 2020-07-16
[Email verification] Redirect to back-end vs front-end URL

franckc:
To fix an issue where if an admin has not deployed their shop yet the email verification flow would fail.


shahthepro:
Thanks @franckc for doing this. I actually added those two parameters, because when redirecting we cannot expect `localStorage.activeShop` to be set always. But, don't think it is that big of a problem. If this works for both deployed and non-deployed shops, I'm fine with it.

franckc:
@shahthepro I did test with both an unpublished and a published shop and it worked fine... I'm not familiar with the code regarding localStorage.activeShop. What's the implication if it is not set?

PR #258, opened 2020-07-17
Add sentry and some bug fixes

shahthepro:


cloudbuild.yaml
```diff
@@ -27,12 +27,27 @@ steps:
...
+          --keyring=origin \
+          --key=cloudbuild
+
   - id: 'Build docker image'
     name: 'gcr.io/cloud-builders/docker'
     entrypoint: '/bin/bash'
     args:
       - '-c'
       - |
+        source $(cat .network).env
```
mikeshultz, line 50: Personally would prefer if we didn't stuff a whole ton of custom build config into a .env(encrypted or not).  I liked @franckc's idea for using trigger substitutions for build args.

devops/Dockerfile
```diff
@@ -4,10 +4,12 @@ ARG ENVKEY
 ARG APP_DIR=/app
 ARG DSHOP_BACKEND_SOURCE="backend"
 ARG DSHOP_DAPP_SOURCE="shop"
+ARG SENTRY_DSN
```
mikeshultz, line 7: May want to configure this with an empty default.  Or at least test that this doesn't require the `SENTRY_DSN` build arg to be defined to build the image.

franckc:
Thanks for working on this sir!
Couple of thoughts:
 - Maybe we can use GCP build [variable substitution](https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values). It would entail splitting the current trigger that does both staging and prod into 2 separate triggers. The staging one would define substitution variables DSN_ENTRY=XXX and ENVIRONMENT=staging; the prod trigger would define substitution variables DSN_ENTRY=XXX and ENVIRONMENT=prod
 - If we are not able to restrict Sentry exceptions capture to only admin pages and it applies also to the buyer facing UI, then from a privacy standpoint I think we will need to do both:
    1. ask permission to the merchant to capture their buyers' crash reports
    2. ask for the buyer approval before sending any crash report

shahthepro:
@mikeshultz @franckc Will fiddle around with variable substitutions on GCP. 

> If we are not able to restrict Sentry exceptions capture to only admin pages

We can actually write a custom transporter for Sentry that ignores any non-admin page issues. Or could move the Error Boundary to admin routes. Both should work fine.

> ask permission to the merchant to capture their buyers' crash reports
> ask for the buyer approval before sending any crash report

For both of these, custom transporter will be more apt, since we'd have more control over the when and how to send.

shop/src/pages/admin/settings/shipping/_Edit.js
```diff
@@ -63,7 +63,10 @@ const EditShippingMethod = ({ onClose, shippingZone }) => {
     ...initialState,
 
     ...shippingZone,
-    amount: shippingZone ? (shippingZone.amount / 100).toFixed() : 0,
+    amount:
```
shahthepro, line 66: @nick There is a merge conflict on this line. Just noticed you changed this line to `amount` to fix the same error on master. We actually store the shipping price in cents, so setting this to just `amount` would break editing a shipping option. Or are we going to start storing in $ instead of cents?

shahthepro: Nvm, didn't notice the lodash get on the top

shahthepro:
@franckc @mikeshultz Moved the SENTRY_DSN to substitution variables. We now have two separate build triggers (one for `master` and one for `stable` branch). 

@nick @franckc Sentry is now optional, Sentry's file are loaded from CDN (and excluded if DSN is not set during build) instead of NPM. There is a shop-level config to disable it. When enabled, errors on admin page would be auto-reported. For user-facing errors, the user would get a prompt to report/skip the error.

Going to merge and test this on Rinkeby. Let me know if there are any comments, will address them

PR #269, opened 2020-07-18
[Listener] Refactor handleLog method so that it gets retried by the queue in case of IPFS failures

### LINKED ISSUE #225: Backend: "IPFS timeout" exception when creating an offer
Seen in the server logs in prod:
2020-07-15T23:14:27.214240084Z [INFO]  dshop.utils.handleLog: Event OfferCreated for listing Id 1-001-233
2020-07-15T23:14:27.535264693Z [INFO]  dshop.utils.handleLog: OfferCreated - 559 by 0xDF73aF150b8E446a6D39FDdc2CFA7Bf067B88936
2020-07-15T23:14:27.535335301Z [INFO]  dshop.utils.handleLog: IPFS Hash: QmZBXbhdAa37sAbCX1h95coAePvuXiwwjn92hbvjeXE3Xp
2020-07-15T23:14:37.539693866Z [ERROR] dshop.utils.handleLog: IPFS gateway timeout

Associated Sentry error: https://sentry.io/organizations/origin-protocol/issues/1786607561/events/de24376de8ac41cab0655d887f5beb12/

Sentry shows the stacktrace as:
```
      statusStr: 'error',
      data: { error: e.message },
      updatedBlock: event.blockNumber,
      createdAt: new Date(event.timestamp * 1000),
      createdBlock: event.blockNumber,
      ipfsHash: event.ipfsHash
    })
    Sentry.captureException(e)
    return order
  }
```

Which is here: https://github.com/OriginProtocol/dshop/blob/master/backend/utils/handleLog.js#L312

franckc:
This should fix #225 

Still working on writing unit tests...

### DIFF
--- a/backend/utils/handleLog.js
+++ b/backend/utils/handleLog.js
@@ -42,7 +42,7 @@ const MarketplaceABI = Marketplace._jsonInterface
  * @returns {Promise<void>}
  * @throws In case of a recoverable error that should be re-tried.
  */
-async function _handleLog({
+async function handleLog({
   networkId,
   contractVersion,
   address,
@@ -107,12 +107,13 @@ async function _handleLog({
  * Refunds a Stripe payment.
  *
  * @param {Object} event: blockchain log.
+ * @param {models.Shop} shop: Shop DB object.
  * @param {models.Order} order: Order DB object.
  * @returns {Promise<null|string>} Returns null or the reason for the Stripe failure.
  * @throws In case of a recoverable error that should be re-tried.
  * @private
  */
-async function _processStripeRefund({ event, order }) {
+async function _processStripeRefund({ event, shop, order }) {
   log.info('Trying to refund Stripe payment')
   // Load the shop configuration.
   const shopConfig = getConfig(shop.config)
@@ -176,7 +177,7 @@ async function _processStripeRefund({ event, order }) {
  * @throws In case of a recoverable error that should be re-tried.
  * @private
  */
-async function _processEventForExistingOrder({ event, order }) {
+async function _processEventForExistingOrder({ event, shop, order }) {
   const eventName = event.eventName
 
   const updatedFields = {
@@ -188,8 +189,8 @@ async function _processEventForExistingOrder({ event, order }) {
     // If it's a Stripe payment, initiate a refund.
     const paymentMethod = get(order, 'data.paymentMethod.id')
     if (paymentMethod === 'stripe') {
-      const refundError = await _processStripeRefund({ event, order })
-      // Store the refund error in the order data.
+      const refundError = await _processStripeRefund({ event, shop, order })
+      // Store the refund error in the order's data JSON.
       updatedFields.data = {
         ...order.data,
         refundError
@@ -206,6 +207,7 @@ async function _processEventForExistingOrder({ event, order }) {
  * Processes a blockchain event for a new order that has not been recorded yet in the system.
  *
  * @param {Object} event: blockchain event.
+ * @param {string} offerId: fully qualified offer id.
  * @param {models.Shop} shop: SHop DB object.
  * @param {boolean} skipEmail: whether to skip sending a notification email to the merchant and buyer.
  * @param {boolean} skipDiscord: whether to skip sending a discord notification.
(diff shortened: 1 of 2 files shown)


backend/models/index.js
```diff
@@ -3,11 +3,20 @@ require('dotenv').config()
...
+
 const basename = path.basename(__filename)
 const db = {}
 
-const SqliteURI = `sqlite:${__dirname}/../db/dshop.db`
-const URI = process.env.DATABASE_URL || SqliteURI
+let URI
+if (process.env.NODE_ENV === 'test') {
+  // In test environment, always use SQLite with storage in a tmp directory.
+  URI = `sqlite:${process.env.TEST_TMP_DIR}/dshop.db`
```
mikeshultz, line 13: hm, I set up the tests to assign `DATABASE_URL`.  did that not work?

backend/test/setup.js
```diff
@@ -55,8 +58,17 @@ before(async function () {
...
 } else {
-  // Shutdown ganache etc if we're not in watch mode and tests are finished.
+  // Shutdown services and clean up if we're not in watch mode and tests are finished.
   after(async function () {
     await shutdownServices()
+
+    console.log(`Cleaning up test data in ${TEST_TMP_DIR}...`)
+    // Note: fs.rmdirSync(TEST_TMP_DIR, { recursive: true }) is still experimental and does not seem to work.
+    await new Promise((resolve, reject) => {
+      execFile('rm', ['-rf', TEST_TMP_DIR], (err) => {
```
mikeshultz, line 68: Can we be a bit more defensive about `rm -rf`?  Also, and env var to maybe disable this could be useful.  There are times where I want to inspect the dir structure.

backend/utils/handleLog.js
```diff
@@ -96,234 +101,296 @@ async function _handleLog({
...
  */
-async function handleLog(params) {
-  try {
-    return _handleLog(params)
-  } catch (e) {
-    Sentry.captureException(e)
-    throw e // Re-throw.
+async function _processStripeRefund({ event, shop, order }) {
+  if (IS_TEST) {
+    return null
```
mikeshultz, line 120: Maybe drop a log here.

backend/package.json
```diff
@@ -15,7 +15,7 @@
     "migration:generate": "sequelize migration:generate --config db/config.js --migrations-path db/migrations",
     "prettier": "prettier --write *.js \"**/*.js\"",
     "prettier:check": "prettier -c *.js \"**/*.js\"",
-    "test": "NODE_ENV=test mocha --timeout 10000 --exit --file test/setup"
+    "test": "NODE_ENV=test TEST_TMP_DIR=/tmp/dshop-test-`date '+%Y-%m-%d-%H:%M:%S'` mocha --timeout 10000 --exit --file test/setup"
```
mikeshultz, line 18: Kind of defeats the purpose of using an env var over a const doesn't it?

backend/test/setup.js
franckc: yes good point. I'll add a check to make sure TEST_TMP_DIR starts with /tmp
Also will add an option or env var to leave the data.

backend/models/index.js
franckc: It took me a while to figure what was going on. All my tests were failing... What happens with mocha is that setup.js gets evaluated AFTER the test files are loaded up. So the tests end up using whatever DATABASE_URL value is defined in your environment. And the overrides of DATABASE_URL in setup.js are not effective...

backend/package.json
franckc: I had to do that in order for models/index.js to use the proper sqlite location. Otherwise it was using the default sqlite URL which is src/backend/db rather that TEST_TMP_DIR
I'm open to any suggestion for simplifying this...

mikeshultz: If `modes/index.js` and the tests pull the temp dir from the same source(probably somewhere in `models/` that would work.  And you could leave the env var open in case someone wanted to manually set it for whatever reason(using it this way in package.json it can't be overridden without altering package.json).  

This isn't a huge deal, env vars in package.json just irk me a little.

backend/utils/handleLog.js
franckc: done

backend/package.json
franckc: Yes that's a better idea. Done. Thank you for the suggestion.

PR #280, opened 2020-07-21
Sentry changes

shahthepro:
- Logs all errors in admin pages by default
- Allows the shop admin to enable/disable logging errors for users

### DIFF
--- a/backend/routes/shops.js
+++ b/backend/routes/shops.js
@@ -1,5 +1,7 @@
 const omit = require('lodash/omit')
 const pick = require('lodash/pick')
+const sortBy = require('lodash/sortBy')
+
 const Stripe = require('stripe')
 const {
   Seller,
@@ -8,6 +10,7 @@ const {
   Network,
   ShopDeployment,
   ShopDeploymentName,
+  Order,
   Sequelize
 } = require('../models')
 const {
@@ -43,6 +46,9 @@ const {
 const printfulSyncProcessor = require('../queues/printfulSyncProcessor')
 
 const log = getLogger('routes.shops')
+
+const dayjs = require('dayjs')
+const { readProductsFile } = require('../utils/products')
 
 async function tryDataDir(dataDir) {
   const hasDir = fs.existsSync(`${DSHOP_CACHE}/${dataDir}`)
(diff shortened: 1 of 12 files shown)


shop/src/components/SentryErrorBoundary.js
```diff
@@ -19,33 +21,58 @@ class ErrorBoundary extends React.Component {
...
+        return null
+      }
+    }
 
-    const pathname = new URL(event.request.url)
-    const isAdminPage = /^#\/(super-)?admin(\/.*)?$/.test(pathname.hash)
-    if (!isAdminPage) {
-      // Ask permission from the user, if it is not in admin page
-      setErrorEvent(event)
+    if (process.env.NODE_ENV === 'development') {
```
franckc, line 53: When running on local is NODE_ENV always set to 'development"? Even like in watch mode? Just asking to be sure.

franckc:
Sorry for all the back-end-forth on this... Here is the latest based on the discord discussion on discord with @gullible1 and @matthewliu: For the beta launch they suggest the following logic:
 - Sentry enabled by default for all admin pages for all merchants.
 - Sentry enabled by default on all shops pages for all buyers except on the Brave store where it should be disabled.
We'll revisit this logic for the public launch.

backend/routes/shops.js
```diff
@@ -806,20 +806,25 @@ module.exports = function (router) {
...
-        const network = await Network.findOne({ where: { active: true } })
-        const netConfig = getConfig(network.config)
-        const { secret } = await registerStripeWebhooks(
-          req.body,
-          existingConfig,
-          netConfig.backendUrl || existingConfig.publicUrl
-        )
-        additionalOpts.stripeWebhookSecret = secret
-      } else if (req.body.stripeWebhookSecret) {
+      if (process.env.NODE_ENV !== 'development') {
```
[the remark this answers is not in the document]
shahthepro: > Also why disabling it exactly? What if we want to test Stripe payment flow on local?

Stripe webhook cannot be registered with localhost. For testing webhooks on local, you have to use stripe CLI and enter the webhook secret manually.

shop/src/components/SentryErrorBoundary.js
shahthepro: Thanks. Changing this to  `!== production`

PR #291, opened 2020-07-22
DNS Verification Fixen

### LINKED ISSUE #282: DNS changes still not propagated after 16 hours
One of our users is having trouble verifying his DNS records after 16 hours.

Screen_Shot_2020-07-22_at_05 56 55

mikeshultz:
Fixes a few issues with the `/domains/verify-dns` endpoint.  Primarily `/api/v0/dns` returns IPFS hashes, not IPNS names, and IPFS hashes are case-sensitive.  Addressed a few other issues while there.

Fixes #282

### DIFF
--- a/backend/routes/domains.js
+++ b/backend/routes/domains.js
@@ -49,8 +49,7 @@ module.exports = function (router) {
               [Op.like]: `%${domain}%`
             }
           },
-          order: [['created_at', 'DESC']],
-          limit: 1
+          order: [['created_at', 'DESC']]
         })
 
         if (!deployment) {
--- a/backend/routes/domains.js
+++ b/backend/routes/domains.js
@@ -1,12 +1,10 @@
 const fetch = require('node-fetch')
-
 const get = require('lodash/get')
 
-const { authSellerAndShop, authRole } = require('./_auth')
-
 const { getLogger } = require('../utils/logger')
-
 const { Op, Network, ShopDeployment } = require('../models')
+
+const { authSellerAndShop, authRole } = require('./_auth')
 
 const log = getLogger('routes.domains')
 
--- a/backend/routes/domains.js
+++ b/backend/routes/domains.js
@@ -5,11 +5,13 @@ const get = require('lodash/get')
 const { authSellerAndShop, authRole } = require('./_auth')
 
 const { getLogger } = require('../utils/logger')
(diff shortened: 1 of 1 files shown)


backend/routes/domains.js
```diff
@@ -21,33 +21,61 @@ module.exports = function (router) {
 
       log.debug('Trying to verify DNS records of domain', domain, hostname)
 
+      // Verify the domain is RFC-valid
+      if (!domain.match(DNS_VALID)) {
+        log.debug(`${domain} is not a valid domain`)
+        return res.send({
+          success: true,
```
franckc, line 28: should this be success:false? same comment for similar code below.

backend/routes/domains.js
```diff
@@ -21,33 +21,61 @@ module.exports = function (router) {
...
 
-        const networkConfig = getConfig(network.config)
+        const deployment = await ShopDeployment.findOne({
+          where: {
+            domain: {
+              [Op.like]: `%${domain}%`
+            }
+          },
+          order: [['created_at', 'DESC']],
+          limit: 1
```
franckc, line 53: are order and limit necessary given findOne is called?

backend/routes/domains.js
```diff
@@ -65,7 +93,7 @@ module.exports = function (router) {
           error: `Your DNS changes haven't propagated yet.`
         })
       } catch (err) {
-        console.error('Failed to check DNS of domain', err)
+        log.error('Failed to check DNS of domain', err)
         res.send({
           error: 'Unknown error occured'
```
franckc, line 98: is this missing success:false?

backend/routes/domains.js
mikeshultz: I'm not trying to alter outputs from what the FE expects here.

backend/routes/domains.js
mikeshultz: Apparently limit is not.  Might be confusing this with another ORM(SqlAlchemy?) that will error if findOne returns more than one record.  Removed.  Order is necessary though since I want the newest only.

backend/routes/domains.js
mikeshultz: Same answer as above.

backend/routes/domains.js
mikeshultz: +1 for a standardized output in general, however.

PR #293, opened 2020-07-23
[Offer Processor] Refactor and improve error handling

### LINKED ISSUE #272: [Offer Processor] Delays in mining can cause order processing errors
As captured in prod by Sentry in [this event](https://sentry.io/organizations/origin-protocol/issues/1792745401/events/3e4f7ead8a4e42138ab0adc38bafde30/), the offer processor may timeout while waiting for a tx to get mined.

Overall, delays in mining could cause multiple problems:
 - since the offer process queue retries, multiple offers for the same order may end up being created. The scenario would be as follow: queue calls offer process to process order A. Processor times out waiting on tx to get mined. Before the retry happens, the first tx does get mined. A 2nd tx for the same order is sent when the retry happens.
 - the offer creation for a subsequent order may fail due to using an invalid nonce. The scenario would be as follow: queue calls offer process to process order A. Mining is slow. While the processor is waiting for tx for order A to get mined, an order B is placed. Offer processor uses the same nonce as for tx A. Only one tx (either A or B) gets mined and the other fails. And so on so forth if new orders are placed at a rapid clip.

franckc:
Refactored and improved the makeOffer processor logic to make it more robust to errors.

 - Record transaction data in the DB. Previously this table was no populated (it currently sits empty in prod).
 - Attempt to recover from crashes/interruptions by using data from the transactions table 
 - Use ether.js rather than web3
 - Block on confirmation rather than timing out
 - Added a gas price multiplier to speed up tx mining. Currently set at 101% of the gas price returned by the node. The purpose is to be slightly ahead compared to other tx submitted to the network with default gas price.
 - Added unit tests

Should fix #272


backend/db/migrations/20200722203947-addColumnsToTransactions.js
```diff
@@ -0,0 +1,44 @@
...
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async () => {
+      await queryInterface.dropTable(tableName) // Note: this table was empty so it's fine to drop it.
+      await queryInterface.createTable(tableName, {
+        id: { type: Sequelize.INTEGER, autoIncrement: true, primaryKey: true },
+        shop_id: Sequelize.INTEGER,
+        network_id: Sequelize.INTEGER,
+        wallet: Sequelize.STRING,
```
mikeshultz, line 14: What is "wallet"?

backend/queues/makeOfferProcessor.js
```diff
@@ -34,48 +57,126 @@ async function processor(job) {
...
+    const network = await _getNetwork(shop.networkId)
     const networkConfig = encConf.getConfig(network.config)
     const shopConfig = encConf.getConfig(shop.config)
 
     queueLog(10, 'Creating offer')
     const lid = ListingID.fromFQLID(shop.listingId)
-    const offer = createOfferJson(lid, encryptedData, paymentCode)
-    const ires = await postOfferIPFS(network, offer)
+    const offer = _createOfferJson(lid, encryptedData, paymentCode)
+    const ipfsHash = await _postOfferIPFS(network, offer)
```
mikeshultz, line 76: What's with underscore-prefixing all these?

backend/queues/makeOfferProcessor.js
```diff
@@ -34,48 +57,126 @@ async function processor(job) {
...
+      log.info(
+        `Found pending transaction ${transaction.id} job ${job.id} hash ${transaction.hash} for wallet ${walletAddress}`
+      )
+
+      // If it is not the transaction from our job. Do not try to recover it.
+      // Let it get recovered by the job that created it.
+      // Fail our job for now, it will get retried.
+      if (transaction.jobId === job.id) {
+        throw new Error(
+          `Pending transaction does not belongs to job ${job.id}. Bailing.`
```
mikeshultz, line 122: What happens for busy shops that say have 5 pending orders?  Does this job just fail repeatedly until the retry limit is reached?

backend/queues/makeOfferProcessor.js
```diff
@@ -34,48 +57,126 @@ async function processor(job) {
...
+        ipfsHash,
+        jobId: job.id.toString()
+      })
+    }
+
+    // Wait for the tx to get mined.
+    // Note: this is blocking with no timeout. Depending on the network conditions,
+    // it could take a long time for a tx to get mined. This will become a bottleneck
+    // in the future when the transaction volume scales up. A potential solution
+    // will be to run the confirmation logic as a separate queue with multiple workers.
```
mikeshultz, line 167: FYI: I think there will be multiple workers.  Right now, should be 3 for prod.

backend/queues/makeOfferProcessor.js
```diff
@@ -34,48 +57,126 @@ async function processor(job) {
...
+    log.info('Waiting for tx confirmation...')
+    confirmation = await _waitForMakeOfferTxConfirmation(marketplace, tx)
+    const { receipt, offerId } = confirmation
+
+    // Update the transaction in the DB.
+    const oid = new OfferID(lid.listingId, offerId, network.networkId)
+    await transaction.update({
+      status: TransactionStatuses.Confirmed,
+      blockNumber: receipt.blockNumber,
+      offerId: oid.toString() // Store the fully qualified offerId.
```
mikeshultz, line 178: Can we guarantee offer ID is unique in the `transaction` table?

backend/utils/createListing.js
```diff
@@ -73,20 +85,29 @@ async function createListing({ network, pk, title }) {
...
-    tx = await Marketplace.methods
-      .createListing(ipfsBytes, 0, account.address)
-      .send({ from: account.address, gas: 350000 })
-  } catch (e) {
-    log.error(e)
-  }
-
-  web3.currentProvider.connection.close()
+  // Send the transaction to the network and wait for it to get mined.
+  const options = { gasLimit: 350000 }
```
mikeshultz, line 89: Second time I've seen this.  Might be good to have gas constants somewhere.

backend/db/migrations/20200722203947-addColumnsToTransactions.js
franckc: bad naming on my part. it's the wallet address the transaction was sent from. let me rename to "from". and might as well add a "to" column in there.

mikeshultz: I might recommend not using `from` as it's a reserved SQL keyword, but something more clear would be great.

backend/queues/makeOfferProcessor.js
franckc: This is to denote them as private. I think it is a fairly common convention? but happy to change if that's unorthodox.

backend/queues/makeOfferProcessor.js
franckc: Good catch. I meant to use the !== operator. I will also add a unit test.
The idea is that the pending transaction will only get processed by the retry for the jobId that created it.
Other transactions will throw exception and get retried and should eventually succeed as part of a retry once the pending transaction has been mined.

backend/queues/makeOfferProcessor.js
franckc: This is predicated on a single worker processing the queue.
Afaik by default Bull's concurrency is set to 1: https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queueprocess

Added a comment to make sure concurrency does not get increased inadvertently in the future.

backend/queues/makeOfferProcessor.js
mikeshultz: Generally think it's private in classes but I really don't care that much.  Fine to leave as is.

backend/queues/makeOfferProcessor.js
mikeshultz: A thrown exception will count as a retry for the job though, wouldn't it?  This could cause it to fail say 3 times before the transaction is actually returned from that query.

backend/queues/makeOfferProcessor.js
mikeshultz: I think that concurrency setting only accounts for workers in that process, not multiple processes.  We currently have 3 instances in prod.  It don't think it will run the same job concurrently, but I'm not sure what you're concerned about, either.  Just wanted to point out that blocking here doesn't necessarily hold up other jobs like your comment suggested.

backend/queues/makeOfferProcessor.js
franckc: In the future we might have more transaction types recorded. For example OfferWithdrawal/Acceptance/Finalization and in that case offerId won't be unique. So I think I'll leave it as is for now.

backend/queues/makeOfferProcessor.js
franckc: I'm so glad we do code reviews! THANK YOU. I was under the incorrect assumption that Bull's concurrency settings are global across node.js processes. They are not. Let me rework a bit the code. I think I can still make it work by checking if there is any pending transaction from the wallet address and if any and it's not the same job id then retry...

backend/db/migrations/20200722203947-addColumnsToTransactions.js
franckc: Good point. Used column names fromAddress and toAddress

backend/utils/createListing.js
franckc: Moved to a common file: packages/src/marketplace.js

backend/queues/makeOfferProcessor.js
franckc: Inspected the logic and it should work for multi-consumer nodes.

franckc:
Merging this in. Will monitor those changes in prod...

PR #294, opened 2020-07-23
DNS Record Suggestions

### LINKED ISSUE #289: Domain roots can not be CNAMEs
Our "Add a Custom Domain" UI currently will allow a user to use a root domain(example.com) and will instruct users to set this as a CNAME record.  [This is not really possible](https://serverfault.com/a/170200/111869) and only tells our users to attempt something that doesn't work.

We probably need to disallow this and instruct users on another way of doing this, or instruct users to set the domain as an A record with our ingress' IP address.

CNAMEs are best for this because we can update our configuration without worrying about users. While I don't expect to lose control of that IP any time soon, if we ever want to change it, it will be a pretty major effort to wrangle all the users to update their records.

mikeshultz:
DNS record suggestions were invalid for root-zone records(e.g. originprotocol.com).  This tries to mitigate that by suggesting the user create an A record for their shop when necessary.  If that IP changes we're in for a ton of hassle but I don't see us dropping that if we have a choice in the near future.  And this is better than preventing users from using root-zone records entirely.

This PR touches both the FE and BE.

Fixes #289

### DIFF
--- a/backend/routes/domains.js
+++ b/backend/routes/domains.js
@@ -118,11 +118,11 @@ module.exports = function (router) {
     }
 
     // A gateway with an unusual port won't work for this
-    /*if (network.ipfs.includes(':')) {
-        log.warn('Invalid IFPS gateway. Cannot use as AutoSSL gateway')
-        success = false
-        error = 'Cannot use this IFPS gateway for DNS configuration'
-      }*/
+    if (network.ipfs.includes(':')) {
+      log.warn('Invalid IFPS gateway. Cannot use as AutoSSL gateway')
+      success = false
+      error = 'Cannot use this IFPS gateway for DNS configuration'
+    }
 
     // If success so far...
     if (success) {
--- a/backend/routes/domains.js
+++ b/backend/routes/domains.js
@@ -2,14 +2,12 @@ const fetch = require('node-fetch')
 const get = require('lodash/get')
 
 const { getLogger } = require('../utils/logger')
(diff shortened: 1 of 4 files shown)


backend/routes/domains.js
```diff
@@ -99,4 +97,60 @@ module.exports = function (router) {
...
+    })
+
+    if (!network) {
+      log.warning('Invalid network provided to /domains/records')
+      success = false
+      error = 'Invalid network'
+    }
+
+    // A gateway with an unusual port won't work for this
+    if (network.ipfs.includes(':')) {
```
franckc, line 121: Dereferencing network will crash if network was not found above.

backend/utils/dns/index.js
```diff
@@ -37,8 +45,131 @@ function isPublicDNSName(v) {
   )
 }
 
+/**
+ * Is the given hostname a valid DNS name
+ *
+ * @param v {string} - The string to check
+ * @returns {boolean} - if the given name is a valid DNS name
+ */
+function isValidDNSName(v) {
```
franckc, line 54: nit: as opposed to duplicating this method's code on the backend and frontend, it could be shared under packages/utils

mikeshultz: I didn't know this existed.  It's a pretty trivial function so I think I'll skip it for this PR but I'll keep this in mind for the future.

backend/routes/domains.js
mikeshultz: Good catch

PR #316, opened 2020-07-28
Manual Payment Method

### LINKED ISSUE #287: Let merchant add manual payment methods
Let merchants accept custom payment methods that are processed outside of their Dshop (e.g. bank transfers, cash on delivery etc).

They should be able to:

- [ ] Click on "Add payment method" under Manual Payment Methods ([see on Zeplin](https://zpl.io/VKR5GDX))
- [ ] Set up their manual payment method by proving a name (e.g. Bank transfer), description that shows up on payments page during checkout (e.g. Pay by bank transfer) and payment instructions that shows up on the order confirmation (e.g. send the payment to this account) - [See on Zeplin ](https://zpl.io/brBj1ve)
- [ ] Go to their Orders page and mark orders as "paid", once they receive the payment ([See on Zeplin](https://zpl.io/bzEJ183))

shahthepro:
Closes #287


shahthepro:
@nick @franckc I'm more or less done with this PR. However, would like to hold off merging until the launch. Don't want to break anything before the launch

franckc:
@shahthepro Nice work on this! Will review. And yes let's hold off until after the launch for merging. Perhaps we could merge on Monday morning along with @nick #329 and do an in-depth testing on staging before we deploy to prod?

shop/src/pages/checkout/ChoosePayment.js
```diff
@@ -58,6 +60,7 @@ const reducer = (state, newState) => ({ ...state, ...newState })
 const ChoosePayment = () => {
   const history = useHistory()
   const { config } = useConfig()
+  const { shopConfig } = useShopConfig()
```
nick, line 63: This won't be available to non-logged in users. The front end should only be reliant on stuff in config.json, so we should probably add in a setting in there to indicate if offline payments are accepted or not

shahthepro: Actually we can move the `offlinePaymentMethods` out of encrypted config to `config.json`. Don't think we have any reason not to do so. Since we are anyhow showing its content in the frontend

shahthepro: Have moved this to the `config.json`. Nothing seems broken on local, gonna merge this to master

PR #327, opened 2020-07-29
Stripe RESTful API CLI script

mikeshultz:
Was using this this week to verify Webhook configurations without actually having access to the users' dashboard.  Figure it might be useful to others.


mikeshultz:
Cool, I missed that.  Will keep it in mind in the future.

PR #329, opened 2020-07-29
Merchant currency select

nick:


backend/routes/exchange-rates.js
```diff
@@ -0,0 +1,51 @@
...
+
+  const response = await fetch(exchangeURL)
+  if (!response.ok) {
+    throw new Error(response.error)
+  }
+  log.info('Fetched exchange rates OK')
+  return await response.json()
+}
+
+const fetchExchangeRates = memoize(fetchExchangeRatesRaw)
```
franckc, line 41: Note (I found out today while looking at another bug on code using memoize): in case of an async function, memoize caches the Promise and therefore potentially the exception returned by the method. Let's say there is a network bleep or the crypto-compare API for some reason returned a temporary error, the memoize cache will return that error for 5min until the cache gets cleared. An alternative could be to have a wrapper function that clears the cache in case fetchExchangeRatesRaw returned an error.
Something like this:
```
const fetchExchangeRatesCache = memoize(fetchExchangeRatesRaw)

async function fetchExchangeRates(target) {
  try {
     fetchExchangeRatesCached(target)
  } catch(e) {
     fetchExchangeRatesCache.cache.delete(target)
     throw e
  }
```

PR #346, opened 2020-07-31
Add PayPal payment method

### LINKED ISSUE #322: Integrate PayPal as a payment option
This is a critical method of payment, especially in regions such as Latin America where payments are tricky and Stripe isn't supported. Using their hosted flow might be easiest.

shahthepro:
Closes #322


backend/routes/shops.js
```diff
@@ -873,6 +877,23 @@ module.exports = function (router) {
         additionalOpts.printfulWebhookSecret = ''
       }
 
+      if (req.body.paypal) {
+        log.info(`Shop ${shopId} - Registering PayPal webhook`)
+        await paypalUtils.deregisterWebhook(shopId, existingConfig)
```
nick, line 882: Can we check the webhook doesn't already exist instead of deleting / recreating every time we PUT /shop/config?

shahthepro:
@nick Have addressed all your comments. Merging this in now.

backend/routes/paypal.js
```diff
@@ -0,0 +1,233 @@
...
+    next()
+  }
+
+  const webhookHandler = async (req, res, next) => {
+    const { shopId } = req.params
+
+    try {
+      const event = req.body
+
+      const externalPayment = await ExternalPayment.create({
```
franckc, line 166: Couple comments:
 - Why calling ExternalPayment.create then afterwards calling save? This results in 2 writes to the database. A single call to create() with all the data should suffice?
 - It would likely be helpful to be able to differentiate between Stripe and PayPal rows when querying the data. So might be good to add a new column "processor" which would be an ENUM with currently 2 possible values: Stripe or PayPal.

backend/queues/makeOfferProcessor.js
```diff
@@ -62,7 +64,7 @@ async function processor(job) {
     job.log(str)
     job.progress(progress)
   }
-  const jobId = `${job.queue.name}-${job.id}` // Prefix with queue name since job ids are not unique across queues.
+  const jobId = `${get(job, 'queue.name', '')}-${job.id}` // Prefix with queue name since job ids are not unique across queues.
```
franckc, line 67: I think this was failing for you because you were using the in-memory queue vs redis.
Sorry I think I added this log.
It might be more elegant to fix it by adding a queue field to the job object here:
https://github.com/OriginProtocol/dshop/blob/master/backend/queues/fallbackQueue.js#L35
Like
```
    const job = {
      queue: this
      data: data,
      progress: () => undefined,
      log: log.info
```

PR #355, opened 2020-08-03
[IPFS get] Do not cache exceptions

franckc:
The back-end processing queue had a few failures caused by IPFS timeouts. All the retries failed with logs that look like this:
```
2020-08-02T04:49:47.198016519Z [INFO]  dshop.utils.handleLog: Event OfferCreated for listing Id 1-001-233
2020-08-02T04:49:47.201167573Z [DEBUG] dshop.utils.events: Upsert event...
2020-08-02T04:49:47.207408834Z [DEBUG] dshop.utils.events: Event exists
2020-08-02T04:49:47.210113302Z [INFO]  dshop.utils.handleLog: OfferCreated - 592 by 0xDF73aF150b8E446a6D39FDdc2CFA7Bf067B88936
2020-08-02T04:49:47.210283391Z [INFO]  dshop.utils.handleLog: IPFS Hash: QmYNkDPPDtX1QbDBrjLSBJwr66GabdDh6DHG6vXwmeYNoG
2020-08-02T04:49:47.215113744Z [INFO]  dshop.utils.handleLog: Using IPFS gateway https://ipfs.ogn.app for fetching offer data
2020-08-02T04:49:47.21534627Z [INFO]  dshop.utils.handleLog: Fetching offer data with hash QmYNkDPPDtX1QbDBrjLSBJwr66GabdDh6DHG6vXwmeYNoG
2020-08-02T04:49:47.21873529Z [ERROR] dshop.queues.eventsProcessor: Log handler failed: IPFS gateway timeout
```

Notice that based on log timestamps, fetching the offer data from IPFS returned with a timeout immediately, despite the timeout being set to 60sec... I checked the content for that hash is available by logging into a server container and fetching the content.

It turns out there is a subtlety in the way memoize works for async functions. It caches the Promise. Therefore, if the Promise gets rejected (due to an IPFS timeout), all subsequent calls to the function for the same IPFS hash would return the same exception as opposed to trying to fetch the content again. Which explains why the queue retries all failed.

Fixed by adding a wrapper method that clears the memoize cache in case of an exception.

Also cleaned up Jsdoc for this file.

### DIFF
--- a/backend/utils/_ipfs.js
+++ b/backend/utils/_ipfs.js
@@ -1,15 +1,18 @@
-const bs58 = require('bs58')
-const FormData = require('form-data')
-const fetch = require('node-fetch')
-const memoize = require('lodash/memoize')
+const {
+  get,
+  getText,
+  post,
+  postBinary,
+  getBytes32FromIpfsHash,
+  getIpfsHashFromBytes32,
+  gatewayUrl
+} = require('@origin/ipfs')
+const { memoizePromise } = require('@origin/utils/memoize')
 
 require('dotenv').config()
+
 const config = require('../config')
-const { getLogger } = require('../utils/logger')
-
 const { DATA_URL, IPFS_GATEWAY } = require('./const')
-
-const log = getLogger('utils.handleLog')
 
 /**
  * Resolve an IPFS gateway from whatever's available.  Either a provided config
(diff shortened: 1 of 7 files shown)


franckc:
> Looks good. Did you just test it manually, or wondering if it's worth adding a simple test in here...

Yes I did test manually but let me add a quick unit test. Good to have!

backend/utils/_ipfs.js
```diff
@@ -186,20 +165,44 @@ async function getTextFn(gateway, hashAsBytes, timeoutMS) {
...
+ *
+ * @param {string} gateway: URL of the IPFS gateway to use.
+ * @param {string} hashAsBytes: IPFS hash.
+ * @param {number} timeoutMS: Timeout in msec.
+ * @returns {Promise<string>}
+ * @throws
+ */
+async function getText(gateway, hashAsBytes, timeoutMS) {
+  try {
+    return getTextCache(gateway, hashAsBytes, timeoutMS)
```
mikeshultz, line 182: I vaguely recall dealing with this before in graphql(and [replacing memoize](https://github.com/OriginProtocol/origin/blob/master/packages/graphql/src/utils/memorize.js)).  Are you sure this will throw as you expect?

franckc: Thank you for pointing this out. You were right, it did not throw an exception. Fixed and added unit test.

franckc:
- Reworked the memoize wrapper and moved it under @origin/utils/memoize
 - Refactored the code a bit so that we now have a single place @origin/ipfs/index where this code leaves as opposed to duplicating it under backend/src/util/_ipfs and @origin/ipfs

packages/utils/package.json
```diff
@@ -16,7 +16,8 @@
   },
   "dependencies": {
     "lodash": "^4.17.19",
-    "logplease": "^1.2.15"
+    "logplease": "^1.2.15",
+    "@origin/utils": "^0.1.0"
```
nick, line 20: Depends on itself?

franckc: Oops. Good catch. That was supposed to be in packages/ipfs/package.json. Fixed.

Issue #345, opened 2020-07-31
Potential missed ListingCreated events


mikeshultz:
Seen a few occurrences of this in prod recently.  I don't know if this is causing issues or is intentional.  Just looks weird to me.

    [INFO]  dshop.utils.handleLog: Event ListingCreated for listing Id 1-001-2465
    [DEBUG] dshop.utils.handleLog: Event is not for a registered dshop. Skipping.

The listing JSON [does appear to be a Dshop listing](https://fs.ogn.app/ipfs/QmfPimnYhWSay6AXYMUnrx4LX3nMHoskbrvetFdEdysbMR):

    {
      "__typename": "UnitListing",
      "schemaId": "https://schema.originprotocol.com/listing_3.0.0.json",
      "listingType": "unit",
      "category": "schema.forSale",
      "subCategory": "schema.clothingAccessories",
      "language": "en-US",
      "description": "Origin DShop Store",
      "media": [],
      "price": {
        "amount": "0",
        "currency": "fiat-USD"
      },
      "amount": "0",
      "currency": "fiat-USD",
      "acceptedTokens": ["token-ETH"],
      "commission": {
        "currency": "OGN",
        "amount": "0"
      },
      "commissionPerUnit": {
        "currency": "OGN",
        "amount": "0"
      },
      "requiresShipping": false,
      "unitsTotal": 1000
    } 

Could be from a custom deployment, which would make this sane and okay.  There's no DB entries matching this listing(may or may not be expected).  The numbers aren't really expected, however.

franckc:
I wonder if this is a race condition:
 - A listing is created for a new shop on the front-end via MetaMask
 - Transaction get mined
 - The back-end listener may pick up the event faster than the front-end
 - In that case the front-end won't have updated listing_id in the shops table, which would explain why the listener does not find it...

The good news is that the logic does not process ListingCreated event (only Offer events are). See [code](https://github.com/OriginProtocol/dshop/blob/master/backend/utils/handleLog.js#L242).

One minor impact though is that we won't record that event in the DB which down the road could be useful. See [code](https://github.com/OriginProtocol/dshop/blob/master/backend/utils/handleLog.js#L97).

franckc:
Now, I checked the DB and I don't see any row in the shops table associated with listing id 1-001-2465.
It's possible the user did not wait for the transaction to be mined before killer their browser tab running the dshop deployer, resulting in the front-end not registering the listing id with the back-end?

[...]

PR #356, opened 2020-08-03
Backend for watching ListingCreated transactions

franckc:
Backend to deal with #345 

- Added a new queue "listingCreated" for processing jobs that wait for a tx to get mined and what it does update the associated shop's listingId
- Added a POST and GET /shop/listing route that can be used by the UI to a) send a txHash to watch and b) know if a listingId is associated yet with the shop

This is WIP. I haven't tested but wanted to send out to get early feedback on the approach.

### DIFF
--- a/backend/queues/listingCreatedProcessor.js
+++ b/backend/queues/listingCreatedProcessor.js
@@ -26,16 +26,14 @@ function attachToQueue() {
 }
 
 /**
- * Records a credit card purchase on the blockchain by making
- * an offer on the marketplace contract.
- * Note: several processors may get started, resulting in
- * multiple jobs getting processed concurrently.
+ * Waits for a blockchain ListingCreated transaction to get confirmed,
+ * then updates the associate shop's ListingId.
  *
  * @param {Object} job: Bull job object.
  * job.data is expected to have the following fields:
- *   txHash
- *   fromAddress
- *   shopId
+ *   {string} txHash: hash of the ListingCreated tx to watch.
+ *   {string} fromAddress: Address that sent the transaction
+ *   {number} shopId: Id of the shop
  * @returns {Promise<null||{receipt: ethers.TransactionReceipt, listingId: number, offerId: number, ipfsHash: string}>}
  * @throws
  */
@@ -109,7 +107,9 @@ async function processor(job) {
       // have been mined while we were waiting for this tx to get mined).
       await shop.load()
       if (shop.listingId) {
-        log.info(`ListingId ${shop.listingId} linked to shop ${shopId} while waiting for ${txHash}.`)
+        log.info(
+          `ListingId ${shop.listingId} linked to shop ${shopId} while waiting for ${txHash}.`
+        )
       } else {
         await shop.update({ listingId: lid })
         log.info(
--- a/backend/queues/listingCreatedProcessor.js
+++ b/backend/queues/listingCreatedProcessor.js
@@ -1,0 +1,189 @@
+const ethers = require('ethers')
+
+const { marketplaceAbi } = require('@origin/utils/marketplace')
+
+const { Network, Shop, Transaction } = require('../models')
+const queues = require('./queues')
+const { ListingID } = require('../utils/id')
+const { getIpfsHashFromBytes32 } = require('../utils/_ipfs')
+const { getLogger } = require('../utils/logger')
+const { IS_TEST } = require('../utils/const')
+const { Sentry } = require('../sentry')
+const { TransactionStatuses } = require('../enums')
+
+const log = getLogger('listingCreatedProcessor')
+
+// Wait for 2 blocks confirmation before considering a tx mined.
+const NUM_BLOCKS_CONFIRMATION = IS_TEST ? 0 : 2
+
+/**
+ * Function to start the queue processing.
+ */
+function attachToQueue() {
+  const queue = queues['listingCreatedQueue']
+  queue.process(processor)
+  queue.resume() // Start if paused
+}
+
+/**
+ * Records a credit card purchase on the blockchain by making
+ * an offer on the marketplace contract.
+ * Note: several processors may get started, resulting in
+ * multiple jobs getting processed concurrently.
+ *
+ * @param {Object} job: Bull job object.
+ * job.data is expected to have the following fields:
+ *   txHash
+ *   fromAddress
+ *   shopId
+ * @returns {Promise<null||{receipt: ethers.TransactionReceipt, listingId: number, offerId: number, ipfsHash: string}>}
+ * @throws
+ */
+async function processor(job) {
+  const queueLog = (progress, str) => {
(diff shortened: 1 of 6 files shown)


backend/queues/listingCreatedProcessor.js
```diff
@@ -0,0 +1,181 @@
...
+  try {
+    // Load the associated shop.
+    const shop = await Shop.findOne({ where: { id: shopId } })
+    if (!shop) {
+      throw new Error(`Failed loading shop with id ${shopId}`)
+    }
+
+    const network = await Network.findOne({
+      where: { networkId: shop.networkId, active: true }
+    })
```
nick, line 62: Perhaps throw if there's no network?

backend/queues/listingCreatedProcessor.js
```diff
@@ -0,0 +1,181 @@
...
+    if (!transaction) {
+      throw new Error(
+        `No transaction found in the DB with hash ${txHash} for shop ${shopId} and wallet ${fromAddress}`
+      )
+    }
+
+    // Load the tx from the blockchain based on its hash.
+    queueLog(25, 'Loading tx from the blockchain')
+    const tx = await provider.getTransaction(txHash)
```
nick, line 85: If it's in the mempool but hasn't been mined yet, this would throw? Clients will get their txHash and send to the backend before it's mined.

backend/queues/listingCreatedProcessor.js
```diff
@@ -0,0 +1,181 @@
...
+ * Note: This is blocking. No timeout is set.
+ *
+ * @param {ethers.Contract} marketplace
+ * @param {ethers.Transaction} tx
+ * @returns {Promise<{receipt: ethers.TransactionReceipt, listingId: number|null, ipfsHash: string||null }>}
+ * @private
+ */
+async function _waitForListingCreatedTxConfirmation(marketplace, tx) {
+  // Wait for the tx to get mined with some blocks confirmation.
+  const receipt = await tx.wait(NUM_BLOCKS_CONFIRMATION)
```
nick, line 153: I'm wondering if we should check for outstanding transactions and confirmations manually each time there's a new block instead of relying on tx.wait? If someone speeds up their transaction and the txHash needs to be updated, might be a pain to clean up these tx.wait promises

backend/queues/listingCreatedProcessor.js
```diff
@@ -0,0 +1,181 @@
...
+        )
+      } else {
+        await shop.update({ listingId: lid })
+        log.info(
+          `Shop ${shopId}: tx ${txHash} mined, updated listingId to ${lid}`
+        )
+      }
+    } else {
+      log.info(`Shop ${shopId}: tx with hash ${txHash} was reverted.`)
+      await transaction.update({
```
nick, line 121: Wonder if we should send to Sentry here so we're aware if this happens

backend/queues/listingCreatedProcessor.js
franckc: yes. will do.

backend/queues/listingCreatedProcessor.js
franckc: I think this returns the transaction that was sent to be mined. So it should not error out if the transaction is still in the mempool. I will double check though.

backend/queues/listingCreatedProcessor.js
franckc: Good idea. Will do.

backend/queues/listingCreatedProcessor.js
nick: Yea I was just thinking that if the user is using MetaMask with whatever provider, that TX might not necessarily be available in the Alchemy mempool by the time this code runs

mikeshultz:
Since this is a draft I won't do a full review(unless that's wanted) but wondering if we should talk about this whole speeding up transactions thing.  Why not instead of the client submitting a txHash, they send us their `from` address.  Then we could wait for a `ListingCreated` event from that address and assign it a store.  There's a  small edge case of a user creating multiple shops in quick succession but since they're generic we can just kind of assign them in the order of shop creating.

Alternatively/additionally, maybe we should include a backend+shop ID in the listing IPFS JSON data.

backend/queues/listingCreatedProcessor.js
franckc: Gotcha. Yes that's definitely a possibility. Let me either add a retry inline or will rely on the queue retry to handle this case.

backend/queues/listingCreatedProcessor.js
franckc: Hmm. Yeah good point about sped up transactions... For that case the approach of submitting the merchant's address and then watching for new events from that address would be more reliable. Let me look at that alternative approach. Thanks.

franckc:
@mikeshultz Yes, I was having an inline discussion with @nick (see above). Agreed having the shop register a from address and then monitoring for ListingCreated events would solve this.

How would the merchant prove ownership of that address to the back-end though? We could have them sign a message but that yet another MetaMask popup which is not ideal. Or we could just punt on that aspect for now since there is not much to be gained by claiming someone else's ListingCreated event for your own shop except annoying them.

mikeshultz:
Do we need them to prove ownership?  Are we doing that now with transaction hashes?

nick:
Right now, the merchant does a PUT to /shop/config with `listingId`. 
Instead, I'm suggesting we PUT the txHash of the createListing call, which the listener then watches for (by checking all incoming marketplace events for a matching txHash rather than calling `tx.wait()`).
If the merchant speeds up the transaction, they just do another PUT with a new txHash which appends to a list of watched txHashes (perhaps just comma separated). That way if the non-spedup transaction gets mined just as the sped-up transaction txHash is sent the the backend, we can still update the listing successfully.

franckc:
Thanks for the feedback, super useful. Here is what I'm thinking:
 a) It might be better to send over the wallet address rather than the txHash. Because in case of a speed up the txHash will change and that means the DApp would have to get somehow notified of the speedup txHash so that it can then send it to the backend. If we have the DApp registering with the back-end the wallet address after the web3 wallet gets connected, then the listener can look for ListingCreated event from shop registered addressed and run logic to link listing Id with the corresponding shop.
 b) Regarding proving ownership of the address, the current WIP PR that uses txHash does not address that. It's a separate problem. The idea of adding a field to the Listing JSON, with a value like <network_id>_<shop_id> is interesting. Only concern is that shop_id may not be unique if in the future if we have 3rd party running other dshop backends so there could be collisions. An alernative would be to generate a random "listingCreationCode", store it in the back-en dwith the shop config and then include it in the listing's JSON. I'd say let's keep things simple to start with. We can add that later. Let's get a) done first.

=> Let me update the PR to implement a)

franckc:
Abandoning. Submitted #361 as a replacement.

PR #361, opened 2020-08-04
Add support for ListingCreated events in the listener

franckc:
Backend changes to deal with #345
Rework of the approach in draft PR #356 (abandoned).

- Add new endpoint /shop/wallet that allows the front-end to associate a wallet address with a shop.
- Update the listener's logic to process ListingCreated events. It checks if the transaction was submitted by a wallet associated with a dshop. If it is, it updates the shop's listingId in the DB.


franckc:
@nick ok if I merge this to master? This should be backward compatible and not cause any breakage in prod in case it gets deployed. Also, if the UI ends up needing some additional back-end changes, I'm happy to do so in a separate PR.

nick:
Sounds good. One more thing I realized is that we'd need to update the config.json with the listingId too...

franckc:
Ah yes, good point re updating config.json - will add that before merging...

franckc:
Added logic to update the shop's config.json and updated the unit tests.

Going to merge this in now.

PR #362, opened 2020-08-04
Update From and ReplyTo Emails

### LINKED ISSUE #315: 'From' and 'Support' Email addresses
Initially we had only `supportEmail` in the config that we used as from address as well as support email for all the emails sent out. When a seller creates a shop, the [`supportEmail` is set to `<shop-token>@ogn.app`](https://github.com/OriginProtocol/dshop/blob/master/backend/routes/shops.js#L467). 

1. If the seller doesn't update this support email and hasn't configured any email server settings, network level fallback would be used to send emails. We use AWS SES config (of ogn.app domain), so the emails would work fine in that case. However, the seller wouldn't have access to the email `shop-name@ogn.app`. 

2. If he updates the support email to something like `shopsupport@originprotocol.com` and doesn't configure email servers, the emails won't go through (since the network level credentials belong to a different domain; if it does, it'd be marked as spam/spoof emails).

If he configures the email server for domain `originprotocol.com` and set his `supportEmail` to `<something>@originprotocol.com`, it would work just fine. But I don't think many users have tried configuring their own email servers yet.

So, I had to add one more param `fromEmail` to config, that would be used as from address for all emails (with a fallback value of `no-reply@ogn.app`). But @gullible1 suggested that we should use `supportEmail` as from address for all emails. This would again introduce the above the two problems. 

Some things that we can do:
- Set `Reply-To` to `supportEmail` and `From` to `fromEmail` for all emails
- Set seller's email (from DB) as `supportEmail` (but not `fromEmail` unless the seller configures any email server)
- Pipe/Forward all emails addressed to `<shop-token>@ogn.app` to seller's email. (Unnecessary if we do the above step. But I wonder how many people have already tried to reply to these emails.)
- ??

shahthepro:
Closes #315 

- Makes `supportEmail` and `storeEmail` mandatory while creating a shop
- Uses `storeEmail` as `To` address when sending emails to seller and `supportEmail` for everything else
- Sets 'ReplyTo' header when using network fallbacks (and To address to `support@ogn.app`)


shahthepro:
@nick @franckc I probably want to test this a bit more. Will merge this tomorrow morning my time.

shop/src/pages/admin/_NewShop.js
```diff
@@ -100,6 +116,23 @@ const AdminNewShop = ({ shouldShow, onClose = () => {} }) => {
...
+            We&apos;ll use this address if we need to contact you about your
+            store.
+          </div>
+        </div>
+        <div className="form-group">
+          <label>Customer email</label>
+          <input {...input('supportEmail')} />
+          {Feedback('supportEmail')}
+          <div className="desc">
+            Your customers will see this address if you email them.
```
franckc, line 133: nit: wonder if we should tweak this copy to clarify. the system actually sends the email automatically when new order are received for example. it's not really the merchant emailing themselves. Maybe @gullible1 can think of a better copy?

franckc:
Agreed it would be good to do a round of testing on staging before deploying to prod.

shop/src/pages/admin/_NewShop.js
gullible1: @franckc good point about us the current automated emails.

@shahthepro can I trouble you to help change it to "Your customers will see this address when receiving emails about their order." please? Thanks (:

shahthepro: Done.

shahthepro:
@franckc Did a few more rounds of testing and also added some unit tests. Everything looks good on local, merging this in now.

PR #366, opened 2020-08-04
Use utf8 encoding for original charset for encrypted config

### LINKED ISSUE #358: Corrupt shop configs
Seems like some of the encrypted shop configs are unable to be `JSON.parse`'ed, possibly due to certain characters like emoji's causing issues during the encryption phase.

Currently corrupt configs:

    node scripts/configCli.js --networkId=1 -a --operation=checkConfig

Dump config as a string:

    node scripts/configCli.js --networkId=1 --shopId=ID --operation=getRaw

mikeshultz:
Fixes #358

Includes some tests.  This will chunk up some shops with utf-8 characters. There's 5 with no deployments and 2 are known broken.  The others can probably just be renamed and they'll be fine.

### DIFF
--- a/backend/test/config.test.js
+++ b/backend/test/config.test.js
@@ -1,0 +1,34 @@
+const crypto = require('crypto')
+const chai = require('chai')
+const expect = chai.expect
+
+const { IV_LENGTH, encrypt, decrypt } = require('../utils/encryptedConfig')
+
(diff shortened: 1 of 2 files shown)


mikeshultz:
@nick FYI, I just created a backup table of `shops` called `shops_backup` on the prod DB just in case.

PR #367, opened 2020-08-04
Install and configure Redis for the deployer's GCP image

franckc:
Modify the image install script to setup redis and configure REDIS_URL so that the deployer uses it.


franckc:
The "apt-get install" command installs a default redis configuration and starts the redis service. So I think we should be good. But I'm testing it to make sure...

franckc:
Merging this in so that I can test an image deployment.

PR #381, opened 2020-08-06
Stripe webhook changes

### LINKED ISSUE #369: Stripe Webhook fixes
The current Stripe Webhook implementation needs a few updates:

1. When connecting Stripe, don't delete all webhooks, as there may be some non-Dshop ones that shouldn't be deleted. Instead, either create a new Dshop webhook or update an existing one if needed (ie if there are changes)
2. Only delete Dshop related webhooks when disconnecting Stripe
3. When clicking 'Validate' in the Stripe modal, check that the appropriate webhooks are configured properly
4. Make sure webhooks are correctly configured and active in the `/pay` route and return an error if not. This is so customers are prevented from being charged if there is no active webhook available to process the order.

shahthepro:
Closes #369

### DIFF
--- a/backend/routes/shops.js
+++ b/backend/routes/shops.js
@@ -2,7 +2,6 @@ const omit = require('lodash/omit')
 const pick = require('lodash/pick')
 const sortBy = require('lodash/sortBy')
 
-const Stripe = require('stripe')
 const {
   Seller,
   Shop,
@@ -37,7 +36,7 @@ const path = require('path')
 const { isHexPrefixed, addHexPrefix } = require('ethereumjs-util')
 
 const { configureShopDNS, deployShop } = require('../utils/deployShop')
-const { DSHOP_CACHE, IS_PROD } = require('../utils/const')
+const { DSHOP_CACHE } = require('../utils/const')
 const { isPublicDNSName } = require('../utils/dns')
 const { getLogger } = require('../utils/logger')
 const {
@@ -53,6 +52,8 @@ const log = getLogger('routes.shops')
 
 const dayjs = require('dayjs')
 const { readProductsFile } = require('../utils/products')
+
+const stripeUtils = require('../utils/stripe')
 
 async function tryDataDir(dataDir) {
   const hasDir = fs.existsSync(`${DSHOP_CACHE}/${dataDir}`)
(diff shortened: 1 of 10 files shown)


backend/routes/shops.js
```diff
@@ -887,26 +811,23 @@ module.exports = function (router) {
...
+        log.info(`Shop ${shopId} - Deregistering Stripe webhook`)
+        await stripeUtils.deregisterWebhooks(req.shop, existingConfig)
+        additionalOpts.stripeWebhookSecret = ''
+        additionalOpts.stripeWebhookHost = ''
+        additionalOpts.stripeBackend = ''
+      } else if (req.body.stripe) {
+        log.info(`Shop ${shopId} - Registering Stripe webhook`)
+        const { secret } = await stripeUtils.registerWebhooks(
+          req.shop,
+          req.body,
```
franckc, line 824: Should we pick which fields from the body are sent to Stripe?

And also before making the call to Stripe, should we double check the fields are valid? We have util methods to do so here: https://github.com/OriginProtocol/dshop/blob/master/packages/utils/stripe.js
I understand the UI may already do this validation but imho it's worth double checking.

backend/routes/shops.js
```diff
@@ -887,26 +811,23 @@ module.exports = function (router) {
...
+          existingConfig,
+          req.body.stripeWebhookHost || netConfig.backendUrl
+        )
+        additionalOpts.stripeWebhookSecret = secret
+        // } else if (req.body.stripeWebhookSecret) {
```
franckc, line 829: clean up

backend/routes/stripe.js
```diff
@@ -62,7 +62,13 @@ module.exports = function (router) {
...
     const web3Pk = shopConfig.web3Pk || networkConfig.web3Pk
 
-    if (!web3Pk || !shopConfig.stripeBackend) {
+    const valid = await stripeUtils.webhookValidation(
+      req.shop,
+      shopConfig,
+      shopConfig.stripeWebhookHost || networkConfig.backendUrl
+    )
+
+    if (!web3Pk || !valid) {
```
franckc, line 71: It would be worth logging the shop id and the reason for the error. To help troubleshooting when a merchant complains his CC payments are not going thru.
Same comment for the route /stripe/check-creds below.

backend/routes/stripe.js
```diff
@@ -62,7 +62,13 @@ module.exports = function (router) {
     const shopConfig = getConfig(req.shop.config)
     const web3Pk = shopConfig.web3Pk || networkConfig.web3Pk
 
-    if (!web3Pk || !shopConfig.stripeBackend) {
+    const valid = await stripeUtils.webhookValidation(
```
franckc, line 65: nit (not for this line but could not put a comment above): since we are touching this route, might as well document it a bit. add a short jsdoc header for this route explaining when it gets called and the argument expected in the body of the request.

backend/routes/stripe.js
```diff
@@ -209,13 +219,30 @@ module.exports = function (router) {
 
         await stripe.customers.list({ limit: 1 })
 
+        if (!backendUrl) {
```
franckc, line 222: Maybe add a comment explaining why stripeWebhookHost would not be defined?

backend/utils/stripe.js
```diff
@@ -1,5 +1,184 @@
...
+    url: webhookURL,
+    enabled_events: ['payment_intent.succeeded'],
+    description: 'Origin Dshop payment processor',
+    metadata: {
+      dshopStore: shop.authToken
+    }
+  }
+}
+
+const validateWebhookData = (webhook, shop, webhookURL) => {
```
franckc, line 19: nit: add jsdoc. same comment for the rest of this file.

backend/utils/stripe.js
```diff
@@ -1,5 +1,184 @@
...
+        await stripe.webhookEndpoints.update(existingWebhook.id, webhookData)
+      }
+
+      return {
+        success: true,
+        secret: stripeWebhookSecret
+      }
+    }
+  } catch (err) {
+    log.error('Failed to register webhooks', err)
```
franckc, line 141: nit: add shop.id to the log

shop/src/pages/admin/settings/payments/StripeModal.js
```diff
@@ -101,11 +102,21 @@ const StripeModal = ({ onClose, initialConfig }) => {
...
+        <>
+          <div className="form-group">
+            <label>Webhook Host Tunnel (To test locally)</label>
+            <input
+              {...input('stripeWebhookHost')}
+              placeholder="https://blahblah.yourtunnel.com"
+            />
+            {Feedback('stripeWebhookHost')}
+          </div>
+          {/* <div className="form-group">
```
franckc, line 114: nit: cleanup dead code

backend/routes/shops.js
shahthepro: Didn't use `pick` since I deserialise the object in that function. Will add the validation

backend/routes/stripe.js
shahthepro: Sure, will do.

backend/routes/stripe.js
shahthepro: Will add that. BTW, `stripeWebhookHost` is only for development environment, since `localhost` cannot be used for webhooks. It should be `networkConfig.backend` on prod.   
    
Could use Stripe CLI for webhooks instead of registering webhooks while running locally, but this way is closer to how things run on prod. Would be easier to identify/reproduce bugs on local, if any.

shahthepro:
Have addressed all your comments @franckc. Merging this in now.

PR #395, opened 2020-08-07
Update shop listing ID via listener. Closes #394

### LINKED ISSUE #394: Update UI to use new back-end route for connecting a shop to a marketplace listingId
A new route /shop/wallet was exposed in the back-end: https://github.com/OriginProtocol/dshop/blob/master/backend/routes/shops.js#L1290

The UI should call it after the user connected their wallet.
Also the UI won't need anymore to watch for the createListing transaction getting mined and send the listing ID to the back-end -  the listener now handles that logic.

Once completed, this should fix #345

nick:


backend/utils/handleLog.js
```diff
@@ -395,7 +396,8 @@ async function _processEventListingCreated({ event }) {
   const shopConfig = JSON.parse(raw.toString())
 
   // Update the config.json listingId field and write it back to disk.
-  shopConfig.listingId = listingId
+  const netPath = `networks[${shop.networkId}]`
+  set(shopConfig, `${netPath}.listingId`, listingId)
```
franckc, line 400: ah good catch! Thanks for fixing this. I did not realize the listingId was stored in the networks object... my bad. I'll adjust the unit tests accordingly.

PR #401, opened 2020-08-10
Updated shipping settings flow

### LINKED ISSUE #286: Improve shipping settings flow
Users are confused by the current shipping settings. It also doesn't allow one to specify where they are shipping from.

We have redesigned the flow ([see on Zeplin](https://zpl.io/brBz3MW)), which allows the user to
- [ ] specify which country they are shipping from
- [ ] specify processing time (for art, food and other items that need "preparation" time)
- [ ] specify shipping time for different countries. Note that we will predefine these (expedited = 1 day, fast = 2 to 7 days, slow = 1+ weeks), which the merchant will select from a dropdown)
- [ ] specify shipping price for each country+shipping time

The buyer should only be able to see options that are relevant to their country.

Processing time should show up under the product description on the product page.

This should also resolve #230 .

shahthepro:
Closes #286


shop/src/utils/useFlattenedShippingZones.js
```diff
@@ -0,0 +1,50 @@
+import { useState, useEffect } from 'react'
+import get from 'lodash/get'
+
+import useShippingZones from './useShippingZones'
+import shippingTimes from './shippingTimes'
+
+const flattenShippingZones = (shippingZones) => {
```
franckc, line 7: nit: add a short comment (jsdoc ideally) explaining what those methods do.

backend/routes/shipping-zones.js
```diff
@@ -1,84 +1,38 @@
...
-
         fs.writeFileSync(
           shippingPath,
-          JSON.stringify(shippingZones, undefined, 2)
+          JSON.stringify(destinations, undefined, 2)
+        )
+
+        const newConfig = setConfig(
+          {
+            ...getConfig(req.shop.config),
```
franckc, line 26: Should we pick the specific shipping related fields coming from the request that need to be persisted?
This is to avoid a situation where somehow the UI sends an invalid config in the request (due to a bug) and the whole shop configuration gets overwritten.

backend/routes/shipping-zones.js
```diff
@@ -1,84 +1,38 @@
...
-        })
-
-        if (!existingZone) {
-          shippingZones.push(newZoneData)
-        }
-
         fs.writeFileSync(
           shippingPath,
-          JSON.stringify(shippingZones, undefined, 2)
+          JSON.stringify(destinations, undefined, 2)
```
franckc, line 21: Is the format of shipping.json changing? If yes, do we need to run a script to update the shipping.json file for legacy shops?

backend/routes/shipping-zones.js
shahthepro: `req.shop.config` is actually from DB. Only the `shippingFrom` and `processingTime` are picked from body

shop/src/utils/useFlattenedShippingZones.js
shahthepro: Will add it

backend/routes/shipping-zones.js
shahthepro: The root structure of the shipping.json is still the same. I did add a `rates` property to them. The UI has conditionally checks for that one. So, don't think we need to update shipping.json for all shops

shahthepro:
@franckc @nick Tried with the older versions of shipping.json file to make sure front end still works with them. Merging this now, but let me know before prod deploy. Would be good to test it out on prod, just in case.

PR #403, opened 2020-08-10
Script to fix Stripe webhooks

franckc:
A recent changes added metadata to the webhooks but legacy webhooks were not updated.

This script goes thru all the shops and verifies if their webhook is properly configured.
If not, it deletes the invalid webhook and creates a new one.


franckc:
Ran the script in prod.

PR #405, opened 2020-08-10
Override config.json when serving from backend

nick:


nick:
Legacy shops, but in general I think the source of truth for those fields should be the backend regardless of what config.json says when serving from the backend

Issue #375, opened 2020-08-05
[Analytics] Make Seller Analytics data available in the DB


franckc:
Implement P0 and P1 from  the [spec](https://docs.google.com/document/d/1KgFl-qD5y0wW-khyQJeZL7M5FWmAYL4e4X4_bsjqIEA/edit).

- [ ] Added new product(s)
- [ ] Number of products added per merchant
- [ ] Activated Stripe
- [ ] Connected web3 wallet

franckc:
Here is what I'm thinking in terms of implementation:
- Have an "ETL" job that runs periodically (probably daily). It would go thru every active shop, extracts the data and write it into dedicated reporting tables in the prod DB.
- The data would get pulled from different sources
   - Existing DB tables. For example shops, shop_deployments, orders
   - Shop configuration stored as encrypted JSON in the DB
   - Shop data on IPFS: products, collections
   - Potentially in the future we could also pull data from other sources such as Stripe, Paypal, Printful.
- The reporting tables would get prefixed with "etl_" to clearly differentiate them vs production tables.
- The ETL job would get scheduled using the Bull framework that we already use for queues

nick:
Sounds good to me 👍

PR #407, opened 2020-08-11
Backend analytics implementation

franckc:
Implement #375 

Implementation details:
 - A recurring daily ETL job is scheduled via the Bull queueing framework.
 - The job goes thru every shop in the system and extracts data either from the DB, the shop's config, it's IPFS deployment or from its staging area
 - Data is persisted in a new table "etl_jobs" which has a flat structure to make it easy to query/analyze. There will be one entry per job per ETL run.


backend/etl/processor.js
```diff
@@ -0,0 +1,215 @@
...
+
+const FETCH_TIMEOUT_MS = 60000 // 60 sec
+
+class EtlJobProcessor {
+  constructor(data) {
+    this.data = data
+  }
+
+  async _fetchJson(shop, filename) {
+    if (shop.published) {
```
nick, line 29: Are we more interested in the 'latest' data, or currently published data? Like if someone creates a shop and publishes it with zero products, then starts adding products over the course of a week or so, we'd be more interested in the unpublished data than the published data?

backend/etl/processor.js
```diff
@@ -0,0 +1,215 @@
...
+    shop.outputDir = `${DSHOP_CACHE}/${shop.auth_token}`
+  }
+
+  /**
+   * Main
+   * @returns {Promise<void>}
+   */
+  async run() {
+    log.info('Running ETL job...')
+    const job = await EtlJob.create({ status: EtlJobStatuses.Running })
```
nick, line 187: Should we make sure a job isn't already running? What happens if we deploy while a job is running?

franckc: Given the jobs are scheduled only every 24hrs and currently a job should take no more than 5 to 10min max to run, I don't expect there would be any running job when a new one kicks off.

In case the daily job gets interrupted, the failure will get recorded in Sentry.
Then we can simply re-run the job from the Bull admin UI.

backend/etl/processor.js
franckc: We take a daily snapshot of the data. There will be one row in EtlShops per shop per day.
So we should be able to track the daily progress of a shop setup.

Before a shop is published, we track products/collections/shippings data on disk, once a shop is published we track that data by downloading it from IPFS.

Makes sense?

backend/queues/etlProcessor.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+}
+
+/**
+ * Adds a recurring task to the queue for running the ETL job.
+ * Note: this only needs to be called once when setting up the ETL pipeline.
+ *
+ * @returns {Promise<string>} The repeatable job id.
+ */
+async function scheduleEtlJob() {
```
shahthepro, line 43: Just curious, when/where do we call this function? What happens in case of a restart of the instance? Will we have to reschedule it? Or does bull stores scheduled jobs to redis too?

backend/etl/processor.js
nick: Makes sense... I'm just wondering if there are downsides to always tracking the data on disk instead of IPFS.

backend/queues/etlProcessor.js
franckc: I'll call this function manually once on staging and prod to setup the job.
That will get persisted by Bull in Redis so we won't need to call it again, unless we wipe out Redis.

backend/etl/processor.js
franckc: So once the shop is published, we always get the data from IPFS. It's only during the initial shop setup before it gets published for the first time that we would be reading the data from the disk. Does that make sense?

nick: It makes sense, just means we won't have visibility into unpublished changes after a shop is deployed. But maybe that's fine 🤷‍♂️

franckc: Got it. I'll ask @gullible1 if that's something she would like. And if yes I can add it in a separate PR.

franckc:
Merging this in to test on staging. Happy to address any further comments as a separate PR.

PR #413, opened 2020-08-12
Product/Variant level SKU and pricing

### LINKED ISSUE #409: Hide product-level pricing/SKU when there are multiple variants
When a user is creating a new product listing and selects "this product has multiple variants", the product-level pricing and SKU should be disabled. Currently users are confused as to which price is being displayed.

The user should then enter the price for each variant type. On the storefront, buyers should see "From $x" (the cheapest variant type) below the product thumbnail. Upon clicking in to the product, the cheapest option should be displayed first.

We will also need to add a "SKU" column to the variants table as it currently doesn't exist.

Screenshot 2020-08-11 at 10 16 41 PM

shahthepro:
Closes #409 

Added `hasVariant` key to `products.json`, otherwise we would need to query both `products.json` and `data.json` (for all filtered products) in the products list page. 

Requires updating product and a deploy before the `From` keyword can appear in the listing page.


backend/utils/products.js
```diff
@@ -95,11 +123,32 @@ function appendToProductsFile(shop, productData) {
     }
   }
 
-  allProducts[existingIndex] = pick(productData, minimalistProductFields)
+  const hasVariants =
```
franckc, line 126: supernit: can also use the Boolean constructor:
const hasVariants = Boolean(productData.variants && productData.variants.length)

shop/src/pages/Product.js
```diff
@@ -37,6 +37,28 @@ function getImageForVariant(productData, variant) {
   }
 }
 
+/**
+ * Finds and returns the variant that has the least price
```
franckc, line 41: nit: the "cheapest" price vs the "least" price

shahthepro: Thanks @franckc, Have to work on my vocabulary.

PR #426, opened 2020-08-13
useShops. Closes #418

### LINKED ISSUE #418: Remove all shops from /auth
Currently the `/auth` endpoint returns every shop a user has access to. Instead, paginate that list and move to its own endpoint.

nick:
Separate list of shops from `/auth` endpoint


backend/routes/shops.js
```diff
@@ -99,25 +98,74 @@ module.exports = function (router) {
...
   )
 
-  router.get('/shop', async (req, res) => {
-    const { sellerId } = req.session
-
-    if (!sellerId) {
-      return res.status(400).json({ success: false })
+  router.get('/shops', authUser, async (req, res) => {
+    const user = req.seller
+    const order = [['createdAt', 'desc']]
```
franckc, line 103: There is no index on created_at so from a perf standpoint it's better to order by id.

PR #442, opened 2020-08-17
Add new endpoint /crypto/payment

franckc:
New route for handling payment to a crypto wallet.

WIP. Left TODO:
 - update makeOfferProcessor
 - add unit test
 - test

### DIFF
--- a/backend/queues/txProcessor.js
+++ b/backend/queues/txProcessor.js
@@ -1,0 +1,125 @@
+const ethers = require('ethers')
+const { Network, Shop, Transaction } = require('../models')
+const queues = require('./queues')
+const { getLogger } = require('../utils/logger')
+const { IS_TEST } = require('../utils/const')
+const { TransactionStatuses } = require('../enums')
+
+const log = getLogger('listingCreatedProcessor')
+
+// Wait for 2 blocks confirmation before considering a tx mined.
+const NUM_BLOCKS_CONFIRMATION = IS_TEST ? 0 : 2
+
+/**
+ * Function to start the queue processing.
+ */
+function attachToQueue() {
+  const queue = queues['txQueue']
+  queue.process(processor)
+  queue.resume() // Start if paused
+}
+
+/**
+ * Waits for a blockchain transaction to get confirmed,
+ * then enqueues a job to make an offer.
+ *
+ * @param {Object} job: Bull job object.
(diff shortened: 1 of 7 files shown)


franckc:
I haven't added unit test yet nor tested but sending this for a first pass review to get feedback on the approach.

backend/routes/crypto.js
```diff
@@ -0,0 +1,185 @@
...
+  /**
+   * Endpoint to request a payment code to the back-end.
+   *
+   * The body of the request is expected to include:
+   *   {string} fromAddress: Address sending the transaction.
+   *   {string} toAddress: Address receiving the transaction.
+   *   {string} amount: transaction amount
+   *   {string} currency: transaction currency (ex: ETH, OGN, etc...).
+   */
+  router.get('/crypto/payment-code', authShop, async (req, res) => {
```
nick, line 35: should be `.post`?

franckc: yes good point since it writes data in the DB. let me update it...

franckc: done

backend/queues/txProcessor.js
```diff
@@ -0,0 +1,135 @@
...
+    throw new Error(`Transaction with hash ${txHash} not found`)
+  }
+  log.info(`Loaded tx with hash ${txHash} from the network`)
+
+  // TODO: check to and from address.
+
+  // Wait for the tx to get mined.
+  queueLog(50, `Waiting for tx ${txHash} to get confirmed`)
+  log.info(`Waiting for tx ${txHash} confirmation...`)
+  const receipt = await tx.wait(NUM_BLOCKS_CONFIRMATION)
```
mikeshultz, line 97: Wondering if this is a good idea for this queue.  It locks up a worker and may timeout and error anyway(unsure).  Perhaps this thing would work better with a high amount of retries and a tuned backoff?

franckc: Yes good point. I'm not too worried about locking up a worker given our current minuscule volume.
But if we scale up that could become a bottleneck. I have added a comment int the code to document this.

franckc:
Added a unit test to verify the functionality. Ready for final review.

nick:
Looked through - all very nice and clean :-)

Will try starting a UI branch to use these new methods...

Issue #180, opened 2020-07-08
Serve UI via IPNS / DNSLink


nick:
Right now the UI and shop data are bundled under a single IPFS hash which means every shop would need to be redeployed in order to get UI updates.

Instead, we could have the UI served via IPNS / DNSLink from, eg, /ipns/ui.ogn.app so that all shops can get UI updates at the same time.

To prevent caching issues we could do something like append the current timestamp to the bundle files (to the nearest 5 mins). eg `Math.floor(+ new Date()/1000 / 60 / 5)`

mikeshultz:
Think I'd like to take a stab at this.  Probably a good time to deal with #120 as well.

Think I'm going to build a CI/CD pipeline to handle this and update the DNS records on the fly.  Thinking we should build a dist package for every release(saved as a cloudbuild artifact if possible), but bundle the last `N` releases in the IPFS directory, which is then served from the ipfs-cluster via DNSLink record for `ui.ogn.app`.

So, it would be like...

- `/ipns/ui.ogn.app/index.html`
- `/ipns/ui.ogn.app/[build_hash1]/app.css`
- `/ipns/ui.ogn.app/[build_hash1]/app.js`
- `/ipns/ui.ogn.app/[build_hash2]/app.css`
- `/ipns/ui.ogn.app/[build_hash2]/app.js`
- ...

Perhaps we could add a file like `builds.json` that the cached app can check to see if it's the latest bundle?  If it's out of date it could toast request a reload by the user.

Thoughts?

mikeshultz:
Thinking about this a bit more.  `index.html` is probably going to have to be part of the store deployment.  That and the config/product data.  The HTML file will be what loads the assets from `ui.ogn.app`.

franckc:
Thanks for writing down your thoughts.
In terms of wanting to bundle the last N releases in the IPFS directory, what are the use cases you are thinking of?
To be able to rollback to an older version?  If it adds complexity I'd be fine only having the latest version.

mikeshultz:
Historical bundles are specifically to deal with #120

nick:
Yea since index.html is the file that'll refer to the JS builds, I'm not sure it can include build hashes. Right now, the index.html for each store is essentially this:

```html
<!doctype html>
<html lang="en" dir="ltr">

<head>
  <title>Dshop</title>
  <link rel="stylesheet" href="app.acb54700.css" crossorigin="anonymous" />
</head>

<body>
  <div id="app"></div>
[...]

PR #454, opened 2020-08-18
Adds pipeline to build and deploy generic UI

mikeshultz:
- Example builds.json: https://storage.cloud.google.com/dshop-ui-staging/builds.json
- Bucket contents: https://console.cloud.google.com/storage/browser/dshop-ui-staging;tab=objects?forceOnBucketsSortingFiltering=false&project=origin-214503&prefix=
- Deployed IPNS contents: https://ui.ogn.app/
- Deployed IPFS contenst: https://ipfs.io/ipfs/QmZ3v5JUmWZByXU8kmJsBDvVELgs2USQJdxkDvmmLHjrm5

(builds.json and bucket contents don't currently match because of my weird testing but they will)

With this proposed solution, an `index.html` build would load builds.json, use the latest hash to load the latest assets.  I haven't built and tested that yet.

Ref #180

### DIFF
--- a/devops/cloudbuild.ui.yaml
+++ b/devops/cloudbuild.ui.yaml
@@ -24,7 +24,7 @@ steps:
         yarn run build
 
         # Get existing build dir
-        node devops/syncBucket.js ${_BUCKET_NAME} ${_BUILD_DIR}
+        node ../devops/syncBucket.js ${_BUCKET_NAME} ${_BUILD_DIR}
 
         # Copy new build files to build dir
         cp public/*app*.js ${_BUILD_DIR}/
--- a/devops/cloudbuild.ui.yaml
+++ b/devops/cloudbuild.ui.yaml
@@ -1,0 +1,74 @@
+---
+substitutions:
+    _BUILD_DIR: /tmp/build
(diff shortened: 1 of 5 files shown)


mikeshultz:
The example IPFS deployment linked above appears to be missing the old builds.  Will need to update the script to sync with the CS bucket before deploy.

nick:
Nice, looking good so far :-)

What is the purpose of the old hashes in builds.json? We don't just need the latest hash?

mikeshultz:
Old hashes in the JSON file are probably not going to be used by the FE.  Unless there's some use case down the road I'm not thinking of.  Those builds should reflect what is available in the ui.ogn.app IPFS directory though, and the build pipeline references it to know what to GC.  

Perhaps the FE may want to know its build age?  Just speculating.

We could make this a little easier for the FE bootstrapper though and just add a text file called `LATEST` or whatever with the latest build hash.  Save a few bytes.

mikeshultz:
Going to merge this after lunch so I can test live cloudbuild.   LMK if you want to see any changes.

nick:
Sounds good. Have the UI piece in #459

PR #459, opened 2020-08-19
UI CDN

nick:
Currently generating a separate file in the `dist` folder, `cdn.html`, so we can test

### DIFF
--- a/shop/public/template-cdn.html
+++ b/shop/public/template-cdn.html
@@ -1,0 +1,74 @@
+<!doctype html>
+<html lang="en" dir="ltr">
+
+<head>
+  <title>TITLE</title>
+  <meta charset="utf-8">
+  <meta name="viewport"
+    content="width=device-width, initial-scale=1, maximum-scale=1, shrink-to-fit=no, user-scalable=no">
+  <meta http-equiv="Content-Language" content="en">
+  <meta name="description" content="META_DESC">
+
+  <% if (htmlWebpackPlugin.options.sentryDSN) { %>
+  <script src="https://browser.sentry-cdn.com/5.19.2/bundle.min.js"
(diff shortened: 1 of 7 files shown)


shop/public/template-cdn.html
```diff
@@ -0,0 +1,74 @@
...
+<script>
+  if (typeof window.openpgp !== 'undefined') {
+    openpgp.config.show_comment = false
+    openpgp.config.show_version = false
+    openpgp.initWorker({ path: 'dist/openpgp.worker.min.js' })
+  }
+  (function () {
+    const uiSrc = 'UI_SRC'
+    window.webpackPublicPath = uiSrc
+    const rnd = +new Date()
```
mikeshultz, line 51: Wonder if this is necessary.  Shouldn't be the end of the world if this is cached.  Though, I don't have any strong arguments for/against here.

shop/public/template-cdn.html
```diff
@@ -0,0 +1,74 @@
...
+  <link rel="data-dir" href="DATA_DIR" />
+  <link rel="icon" href="FAVICON" />
+</head>
+
+<body>
+  <div id="app">
+    <div class="loading-screen"></div>
+  </div>
+</body>
+<script crossorigin src="dist/openpgp.min.js"></script>
```
mikeshultz, line 41: Could point these at ui.ogn.app/ui.staging.ogn.app, as well

nick: I tried that originally but web workers can't work cross-domain

mikeshultz: Could you elaborate?  Not really familiar.  Is web workers functionality that openpgp uses, I guess?

nick: Yea it's just for the openpgp webworker stuff, which it relies on. I haven't looked into it much yet, there might be a workaround, but this was the quickest fix for now

PR #471, opened 2020-08-24
Use EC2 instance ID for the initial super user password

mikeshultz:
n/t

### DIFF
--- a/backend/routes/auth.js
+++ b/backend/routes/auth.js
@@ -45,6 +45,20 @@ async function getGcpSuperAdminPassword() {
   return password
 }
 
+/**
+ * Gets the EC2 instance ID
+ */
+async function getEC2InstanceID() {
+  const url = 'http://169.254.169.254/latest/meta-data/instance-id'
+  const res = await fetch(url)
+  if (!res.ok) {
+    log.error(`GET ${url} returned ${res}`)
+    throw new Error(`Failed fetching EC2 instance ID.`)
+  }
+  const password = await res.text()
+  return password
+}
+
 module.exports = function (router) {
   router.get('/auth', async (req, res) => {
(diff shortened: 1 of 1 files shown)


backend/routes/auth.js
```diff
@@ -45,6 +45,20 @@ async function getGcpSuperAdminPassword() {
   return password
 }
 
+/**
+ * Gets the EC2 instance ID
+ */
+async function getEC2InstanceID() {
+  const url = 'http://169.254.169.254/latest/meta-data/instance-id'
```
franckc, line 52: is that IP guaranteed to never change?

backend/routes/auth.js
```diff
@@ -262,6 +277,30 @@ module.exports = function (router) {
       }
     }
 
+    if (process.env.AWS_MARKETPLACE_DEPLOYMENT) {
+      /**
+       * In the case of a AWS marketplace deployment, the super-admin password
+       * is the EC2 instance ID.
```
franckc, line 283: I'm not familiar with EC2. I assume there is no possibility an attacker could guess the instance ID?

backend/routes/auth.js
mikeshultz: From what I can tell, yes

backend/routes/auth.js
mikeshultz: Not that I'm aware of.  Though I do think these should only be temporary passwords(GCP included).  Perhaps that's something we can fix forward.

franckc: Yep agreed as a separate feature we should encourage (or maybe force) the admin to provide a new password the first time they login.

mikeshultz: See #473

PR #478, opened 2020-08-24
100% discounts

nick:
Support for 100% off discount codes


backend/routes/offline-payment.js
```diff
@@ -8,6 +8,38 @@ const { getConfig } = require('../utils/encryptedConfig')
 const makeOffer = require('./_makeOffer')
 
 module.exports = function (router) {
+  router.post(
```
franckc, line 11: nit: the naming makes it pretty obvious but perhaps still document this route to explain when it gets called by the UI

backend/routes/offline-payment.js
```diff
@@ -8,6 +8,38 @@ const { getConfig } = require('../utils/encryptedConfig')
...
+
+      if (!web3Pk) {
+        return res.status(400).send({
+          success: false,
+          message: 'Error with payment server'
+        })
+      }
+
+      if (!encryptedData) {
+        return res.json({ success: false, message: 'Missing order data' })
```
franckc, line 32: I'm not sure if we have standardize on format for error returned by the back-end yet.But should that return something else than 200? Like a 400 (Bad Request) given that an expected arg is missing?

PR #481, opened 2020-08-25
Adds Route 53 support to backend

mikeshultz:
This adds AWS Route53 support for shop DNS configuration to the backend.  Will followup with the UI update for GCP credentials.

### DIFF
--- a/backend/utils/dns/route53.js
+++ b/backend/utils/dns/route53.js
@@ -1,0 +1,250 @@
+/**
+ * DNS utilities for AWS Route 53
+ */
+const Route53 = require('aws-sdk/clients/route53')
+
+const { append } = require('../../utils/string')
+const { getLogger } = require('../../utils/logger')
+
+const log = getLogger('utils.dns.route53')
+
+let CACHED_CLIENT
+const DEFAULT_TTL = 60 // 1 minute
+const AWS_API_VERSION = '2013-04-01'
+
+/**
+ * Return a Google Cloud DNS API client
+ *
+ * @returns {DNS}
+ */
+function getClient(credentials) {
+  if (CACHED_CLIENT) return CACHED_CLIENT
+  if (!credentials) throw new Error('Must supply GCP credentails')
+  if (typeof credentials === 'string') credentials = JSON.parse(credentials)
+
+  CACHED_CLIENT = new Route53({ apiVersion: AWS_API_VERSION })
+
+  return CACHED_CLIENT
+}
+
+/**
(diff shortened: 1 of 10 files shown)


backend/utils/dns/route53.js
```diff
@@ -0,0 +1,250 @@
+/**
+ * DNS utilities for AWS Route 53
+ */
+const Route53 = require('aws-sdk/clients/route53')
+
+const { append } = require('../../utils/string')
```
nick, line 6: can we just `require('../string')`?

backend/utils/dns/route53.js
```diff
@@ -0,0 +1,250 @@
...
+let CACHED_CLIENT
+const DEFAULT_TTL = 60 // 1 minute
+const AWS_API_VERSION = '2013-04-01'
+
+/**
+ * Return a Google Cloud DNS API client
+ *
+ * @returns {DNS}
+ */
+function getClient(credentials) {
```
nick, line 20: Does this mean changing credentials via the network admin requires a server restart? Perhaps we could memoize it using the JSON.stringified credentials object instead?

backend/utils/dns/route53.js
mikeshultz: :rofl: The other DNS modules use this too.  No idea what happened here.  Thanks

backend/utils/dns/route53.js
mikeshultz: Guess it does.  That is the same behavior that's in the DNS modules deployed right now.  I can pull this cached logic with this PR if you think it's important or we can create an issue and fix forward.  Might be good to use a hash of the credentials to know when to throw away the cached clients.

Not really sure why you suggest memoizing the creds?  Probably makes more sense to cache the inited client which might include further metadata or even connections(speculating), rather than potentially recreating the client multiple times a session.

nick: I meant memoize the getClient function using the hashed credentials as the cache key, similar to what you also suggested. Probably worth making part of this PR as it's a small change.

mikeshultz: Ah, gotcha.  Yeah, misunderstood.  Will see about it.

mikeshultz: Done

mikeshultz:
@nick Added the AWS keys to the form field.  Simple enough I think but would appreciate if you took a quick glance at it to make sure I didn't miss anything.

nick:
LGTM

Issue #303, opened 2020-07-27
Links to shops with chinese characters/emojis in shop name don't work


gullible1:
Some users have tried creating shops with Chinese characters or emojis in their shop names. The links to these stores via the super admin don't work.

franckc:
Deployed change to restrict shop names to alphanumeric characters.

@gullible1 Going to downgrade this to a P2 and I think we should work on it as part of the localization project which is on our Q3 roadmap.

gullible1:
@shahthepro this is related to #288  - in addition to the links to these stores via the superadmin not working, I'm not sure if there might be other issues caused by the Chinese characters.

shahthepro:
Just deployed a fix and tested it on prod. These stores work fine now.

PR #485, opened 2020-08-25
Encode authToken before sending it from client

shahthepro:
Part of #303, encodes `authToken` in request headers in client side and decodes it on the server side middleware.

### DIFF
--- a/backend/routes/_auth.js
+++ b/backend/routes/_auth.js
@@ -33,7 +33,9 @@ async function authSellerAndShop(req, res, next) {
     return res.status(401).json({ success: false, message: 'Not logged in' })
   }
 
-  const authToken = String(req.headers.authorization).split(' ')[1]
+  const authToken = decodeURIComponent(
+    String(req.headers.authorization).split(' ')[1]
+  )
   if (!authToken) {
     return res.status(401).json({ success: false, message: 'No auth token' })
   }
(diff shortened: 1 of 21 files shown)


franckc:
A few naive questions/comments:
 - what issue does this solve? Is it in case the shop's name had some non-ascii characters?
 - also maybe it's just me but the name "authToken" always confuses me. As far as I understand, this is not a credential but just a way to identify which shop to operate on. maybe we should consider renaming to shopIdentifier or equivalent?
 - why can't we just use the shop id? Wouldn't that be simpler?
 - do we need to worry about this breaking existing shops Front-End until the merchant publishes them again?

shahthepro:
> what issue does this solve? Is it in case the shop's name had some non-ascii characters?  

Yes, some shop name's had emojis and some shop name's had non-Ascii characters. The request header had to be encoded before it can be sent.

> also maybe it's just me but the name "authToken" always confuses me. As far as I understand, this is not a credential but just a way to identify which shop to operate on. maybe we should consider renaming to shopIdentifier or equivalent?

Yep, we should change that. I'm not sure why it is called authToken either. What do you think @nick?

> why can't we just use the shop id? Wouldn't that be simpler?

I'm not sure why the name is being used to identify the shop, and yeah I don't see a reason to use `authToken` over `id`. Perhaps @nick can shed some lights here.

> do we need to worry about this breaking existing shops Front-End until the merchant publishes them again?

No, it won't break. decoding ASCII characters should return whatever the input is, as-is.

PR #507, opened 2020-08-26
Option to force new Printful images

nick:


nick:
@shahthepro I reverted your changes to Printful `writeProductData` for collections... can we only run that code if there are no collections initially, and keep it how it is otherwise?

shahthepro:
@nick https://github.com/OriginProtocol/dshop/pull/511 should take care of that

PR #508, opened 2020-08-26
UI for direct crypto payments

nick:


backend/queues/txProcessor.js
```diff
@@ -68,9 +68,11 @@ async function processor(job) {
 
   // Load the transaction from the DB.
   const transaction = await Transaction.findOne({
-    networkId: network.networkId,
-    shopId,
-    hash: txHash
+    where: {
```
franckc, line 71: wow. thanks for catching this!!! interesting that the unit test did not catch it... I guess during testing sqlite must have alway been returning the latest row in the transaction table...

backend/routes/shops.js
```diff
@@ -885,8 +886,12 @@ module.exports = function (router) {
 
         // Update the public and dataUrl in the shop's config to
         // reflect a possible change to the hostname.
-        additionalOpts.publicUrl = getShopPublicUrl(req.shop, netConfig)
-        additionalOpts.dataUrl = getShopDataUrl(req.shop, netConfig)
+        if (!req.body.publicUrl) {
```
franckc, line 889: Why is this check necessary? 
If the admin FE sends a hostname, shouldn't we always update the public and data URL?

I'm thinking of a scenario where:
1. Config has an existing hostname, publicUrl and dataUrl.
2. Merchant updates hostname and saves config
3. Admin FE call BE with hostname, publicUrl and dataUrl set in req.body
=> that would result in hostname getting updated in the config but not dataUrl and publicUrl, which would then cause later on failures in the back-end to fetch data.

backend/routes/shops.js
```diff
@@ -800,6 +800,7 @@ module.exports = function (router) {
         'offlinePaymentMethods',
         'supportEmail',
         'upholdClient',
+        'useEscrow',
```
franckc, line 803: Initially when we deploy this PR, no merchant will have this config set. Which means it will default to undefined false. And therefore the marketplace won't be used for purchases on their shop.

The alternative would be to set useEscrow to true on all existing shops using a script. Not worth it in my opinion, since I haven't heard any merchants saying they want escrow. But just mentioning it...

backend/routes/shops.js
nick: The idea was that if the FE sends publicUrl or daraUrl, those values should be trusted. I needed to override them when doing testing locally so needed this change. Right now the only place publicUrl and dataUrl get sent from the FE is from a super-user page so this should affect regular merchants

backend/routes/shops.js
nick: Yea I think we default to not using escrow even for existing shops.

For 475 perhaps we can use a network level config for deciding whether to record off-chain payments to the marketplace contract? It should probably apply globally rather than per shop

backend/routes/shops.js
franckc: Sounds good. Perhaps add a comment so we don't get confused in the future when looking at this code.

PR #511, opened 2020-08-27
Write collections only if doesn't exist during sync

shahthepro:


nick:
Just to confirm... this should still write the collections using the 'old' method (ie putting everything in the 'Other' category) by default. It should only use the new method (using collections names from Printful) if collections don't already exist

shahthepro:
@nick I had it write `collections.json` only if it was previously empty. Have changed it like this:
1. If `collections.json` is empty, use Printful types as collections
2. If not, push new product Ids to `Other` collection.

Issue #544, opened 2020-09-11
Buyer side order page is broken due to off-chain changes


franckc:
The order page exposed to the buyer is broken due to off-chain changes.
This page relies on loading event from the marketplace for loading the data and that fails since we are no longer using the marketplace contract.

Relevant code:
https://github.com/OriginProtocol/dshop/blob/master/shop/src/utils/useOrigin.js#L113

This issue is currently only relevant on staging since the off-chain changes haven't been deployed yet to staging.

franckc:
This is fixed.

PR #545, opened 2020-09-14
Fix order email link

franckc:
Fix for bug #544 

If the order is off-chain (e.g. order.offerId is undefined), the order page presented to the buyer should load its data directly from the encrypted IPFS blob.

Note: also includes an unrelated small fix to the offchainOrdersMigration.js script.

### DIFF
--- a/backend/utils/emails/newOrder.js
+++ b/backend/utils/emails/newOrder.js
@@ -30,6 +30,7 @@ function optionsForItem(item) {
 
 async function sendNewOrderEmail({
   orderId,
+  order,
   shop,
   network,
   cart,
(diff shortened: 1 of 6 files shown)


backend/utils/emails/newOrder.js
```diff
@@ -113,6 +115,16 @@ async function sendNewOrderEmail({
...
 
+  // Generate the link for the buyer to see their order.
+  // If the order was recorded on the marketplace, the link uses the transaction hash.
+  // Otherwise the link uses the IPFS hash of the encrypted data.
+  const orderUrl = order.offerId
+    ? `${publicURL}/order/${cart.tx}?auth=${cart.dataKey}`
+    : `${publicURL}/order/${order.encryptedIpfsHash}?auth=${cart.dataKey}`
+
+  // Link for the merchant to the orders admin page.
+  const orderUrlAdmin = `${networkConfig.backendUrl}/admin/orders/${orderId}`
```
nick, line 126: this might need to be `${networkConfig.backendUrl}/#/admin/orders/${orderId}` (with an extra `#/`)

nick (approved):

backend/utils/emails/newOrder.js
franckc: ah yes, good catch!

PR #566, opened 2020-09-24
Case insensitive role checks

shahthepro:
The roles are stored in different cases in DB. We should normalize them with a script sooner or later.


franckc:
Nice catch! LGTM

I think we should use ENUM as much as possible vs STRING in the DB. This way there is no possible confusion.

PR #567, opened 2020-09-24
Fix #562

### LINKED ISSUE #562: Order confirmations for Printful products show "free shipping"
On the pundi-x.ogn.app shop, which is integrated with Printful, all of the orders that have been made are showing "free shipping". We should be using the Printful shipping API and charging the customer for shipping, so that the merchant will not have to cover shipping costs.

Screenshot 2020-09-24 at 1 14 05 AM

Screenshot 2020-09-24 at 1 13 54 AM

shahthepro:
`recipient` sent to the shipping API endpoint was sometimes `undefined` if config takes sometime loading. This surfaces that error instead of showing "Free shipping"

### DIFF
--- a/shop/src/pages/checkout/Shipping.js
+++ b/shop/src/pages/checkout/Shipping.js
@@ -19,7 +19,11 @@ function isActive(zone, cart) {
 const CheckoutShipping = () => {
   const { config } = useConfig()
   const [{ cart }, dispatch] = useStateValue()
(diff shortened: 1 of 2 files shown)


nick:
Hmm, should we not just ensure the config is fully loaded?

shahthepro:
@nick Have added that logic

franckc (approved):
LGTM

PR #569, opened 2020-09-24
Do not auto-fulfill orders with pending payment status

franckc:


franckc:
Merging this in - will address any feedback request forward.

PR #588, opened 2020-09-30
Printful related changes

shahthepro:
- [x] Add sale taxes
- [x] Warn before sync for manually managed products
- [x] Fix shipping defaulting to "Free Shipping"

### DIFF
--- a/backend/utils/discounts.js
+++ b/backend/utils/discounts.js
@@ -182,6 +182,8 @@ const validateDiscountOnOrder = async (
   }, 0)
 
   const shipping = get(cart, 'shipping.amount', 0)
+  const taxRate = parseFloat(get(cart, 'taxRate', 0))
+  const totalTaxes = Math.ceil(taxRate * subTotal)
 
   let discount = 0
   if (discountObj.discountType === 'percentage') {
@@ -193,7 +195,7 @@ const validateDiscountOnOrder = async (
 
   const donation = get(cart, 'donation', 0)
 
-  const calculatedTotal = subTotal + shipping - discount + donation
+  const calculatedTotal = subTotal + shipping - discount + donation + totalTaxes
 
   if (cart.total !== calculatedTotal) {
     // Something has gone wrong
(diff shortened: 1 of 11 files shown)


shahthepro:
@nick Merging this now, happy to address any comments in subsequent PRs.

backend/utils/printful.js
```diff
@@ -193,6 +194,50 @@ const fetchShippingEstimate = async (apiKey, data) => {
...
+  })
+  const json = await shippingRatesResponse.json()
+  if (json.result && Number.isFinite(json.result.rate)) {
+    return {
+      success: true,
+      ...pick(json.result, ['rate', 'required'])
+    }
+  } else {
+    return {
+      success: false
```
franckc, line 236: should we log an error? and maybe also capture in Sentry so that we get some visibility when/if this starts failing?

backend/utils/discounts.js
```diff
@@ -182,6 +182,8 @@ const validateDiscountOnOrder = async (
   }, 0)
 
   const shipping = get(cart, 'shipping.amount', 0)
+  const taxRate = parseFloat(get(cart, 'taxRate', 0))
+  const totalTaxes = Math.ceil(taxRate * subTotal)
```
franckc, line 186: Sadly I don't think we have any unit test for calculating discounts. Would be great to add some :)

Another suggestion would be to keep cleaning the utils directory by moving this to logic/discount - really imho utils should only be for low level utilities.

franckc (commented):
Couple comments inline otherwise LGTM

backend/utils/discounts.js
franckc, line 186: Math.ceil will round up to nearest integer. I think we should keep at least 2 decimals?

shahthepro: Will add them

shahthepro: We store everything in cents in DB and localStorage. The UI converts it into dollar before showing

PR #593, opened 2020-10-02
Cleanup shop creation and configs

franckc:
Various cleanups related to the shop creation code:
 - moved bulk of the code to src/backend/logic/shop/create.js; removed dead code; added comments
 - moved shop templates from src/backend/db/templates to src/config/templates
 - moved src/backend/util/config.js to src/backend/config/baseConfig.js and updated it so that it uses the new IPFS cluster rather then the old DApp IPFS node
 - cleanup up back-end test fixtures: creation and login of super-admin user is now done as part of the tests initialization in setup.js (previously it was duplicated in the Shops and Offers tests suite)

Added a new option "checkPgp" to the configCli tool - this is sort of unrelated to this PR but is on this branch since that's when I started working on it that I realized our shop config/creation logic needed some clean up!...


franckc:
@nick @mikeshultz @shahthepro Lmk if you have any feedback otherwise I'll merge this in today and test it on staging before doing a prod deploy...

backend/logic/shop/create.js
```diff
@@ -0,0 +1,351 @@
...
+async function createShop({
+  seller,
+  name,
+  dataDir,
+  supportEmail,
+  web3Pk,
+  listingId,
+  printfulApi,
+  shopType
+}) {
```
mikeshultz, line 129: No objection to anything going on here, but for clarity, I might recommend breaking this down even further into more component functions.  Especially the config mergeing/building going on below.

Not a deal breaker but I think it would help with clarity.

nick (approved):
Awesome... LGTM. Thanks for the cleanup :)

backend/logic/shop/create.js
```diff
@@ -0,0 +1,351 @@
...
+  log.debug('createShop called')
+  shopType = shopType || 'empty'
+  dataDir = kebabCase(dataDir)
```
mikeshultz, line 132: Do we want to entirely trust `kekbabCase` to do our security-escaping?  Since this goes into a CLI command later on, it makes me a little nervous.  That said, I did some brief tests and it does appear to strip the non-alphanumeric filesystem characters I tried.

mikeshultz (approved):
Some comments inline, but LGTM.

backend/logic/shop/create.js
franckc: Yes agreed, this function is quite long. Let me land this PR and I'll keep working on better modularity.

backend/logic/shop/create.js
franckc: hmm. I like you being security minded. The createShopInDB function checks for any non-alphanumeric name. But that check should be performed higher up, in the createShop method. Let me make this change.

PR #597, opened 2020-10-05
Manual Taxes

### LINKED ISSUE #590: Allow users to manually set tax amounts
Merchants have to adhere to various tax laws in different countries/states and there is currently no way for them to manage their tax settings on Dshop. [See on Zeplin](https://zpl.io/2yreDpL).

We can implement something basic first that allows merchants to:

- [ ] state whether they want taxes to be included in the product price

- [ ] select the countries that require tax

- [ ] if it's the US, the merchant will also be able to select specific states, if needed

- [ ] manually enter a % amount 

On the buyer-facing product page, the buyer will see "Taxes included." displayed below the product price, if the merchant had elected to include taxes in the product price.

shahthepro:
Closes #590

### DIFF
--- a/shop/src/pages/admin/settings/checkout/_Taxes.js
+++ b/shop/src/pages/admin/settings/checkout/_Taxes.js
@@ -1,0 +1,96 @@
+import React, { useEffect, useState } from 'react'
+import fbt from 'fbt'
+import get from 'lodash/get'
+
+import CountryTaxEntry from './_CountryTaxEntry'
+
+const ManualTaxes = ({ state, setState }) => {
+  const taxRates = get(state, 'taxRates', [{
+    rate: 0
+  }])
+
+  return (
+    <div className="manual-taxes">
+
+      <div className="form-group">
+        <label className="mb-0">
+          <fbt desc="Taxes">Taxes</fbt>
+        </label>
+        <div className="desc mb-3">
+          <fbt desc="admin.settings.checkout.taxesDesc">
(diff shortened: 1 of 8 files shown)


shahthepro:
Merging this now, happy to address any comments later

shop/src/pages/admin/settings/checkout/_Taxes.js
```diff
@@ -0,0 +1,110 @@
...
+        </label>
+        <div className="desc mb-3">
+          <fbt desc="admin.settings.checkout.taxesDesc">
+            Charge taxes on your orders. By enabling this features, you
+            acknowledge that it is up to you and a tax professional to determine
+            when and how much tax to collect and remit, and if this feature is
+            right for you.
+          </fbt>
+        </div>
+        {/* <label className="form-check mb-3 font-weight-normal">
```
franckc, line 31: is this commented out intentionally?

franckc (commented):
LGTM

shop/src/pages/admin/settings/checkout/_Taxes.js
shahthepro: Yes, we actually don't store the user's address until the later steps of checkout (which also gets cleared after placing the order), so we cannot show the product prices with taxes. I've raised this with @gullible1, will have to come back to this once a decision is taken regarding this.

franckc: Sounds good. Maybe add a comment in there so that when we look at this code again we have the context as to why it was commented out.

shahthepro: Sure, will do.

PR #608, opened 2020-10-06
Fix paymentType bug for on-chain transactions

franckc:
A new paymentType argument was added to the processNewOrder logic method but in case of an on-chain transaction (e.g. marketplace transaction), it was not passed. Causing the paymentType column to remain blank for those orders in the DB.


shahthepro (approved):
LGTM

PR #609, opened 2020-10-06
Record admin activity

franckc:
This will be used for recording a history of admin activity, including config changes.
It adds a new table "admin_logs" where we can record admin actions.

I refactored the config update logic and at the end inserted a new row in the admin_logs table. It captures the previous shop data (snapshot + list of keys that were changed).

In a separate PR I'll add capturing the admin activity for other actions such store publish and user add/edit/delete.


backend/db/migrations/20201005202813-addAdminLog.js
```diff
@@ -0,0 +1,32 @@
...
+          type: Sequelize.INTEGER,
+          autoIncrement: true,
+          primaryKey: true
+        },
+        action: Sequelize.ENUM(AdminLogActions),
+        shop_id: Sequelize.INTEGER,
+        seller_id: Sequelize.INTEGER,
+        data: isPostgres ? Sequelize.JSONB : Sequelize.JSON,
+        created_at: Sequelize.DATE,
+        updated_at: Sequelize.DATE
```
mikeshultz, line 21: Shouldn't this be append-only?

mikeshultz (commented):
What about separating the audit log from the config versioning?  I wonder if there's just too much going on in that single table right now.

Will take a further look with fresh eyes in the morning.  The mega config merging stuff is confusing.

backend/db/migrations/20201005202813-addAdminLog.js
franckc: Yes, it should never get updated. I'll remove the updated_at.

franckc:
> What about separating the audit log from the config versioning? I wonder if there's just too much going on in that single table right now.

Yes, that's easy for me to do if you guys think it will be more clear.
Like we can have another tables shop_logs to store the config versions, while we keep admin_logs just for admin activity.

> 
> Will take a further look with fresh eyes in the morning. The mega config merging stuff is confusing.

Yes, took me a some time to untangle what was going on. Note that I did not change the update logic at all. I just refactored it a bit. I think we could improve it in the future. My goal for this PR is primarily to clean up/comment/move things to logic/ to make it easier to make changes in the future.

nick (approved):

mikeshultz:
> Like we can have another tables shop_logs to store the config versions, while we keep admin_logs just for admin activity.

I don't think `shop_logs` is any more self-explanatory/self-documenting than `admin_logs` so may as well keep it as-is. And ignoring the config logic then this PR fine by me.  :+1:

franckc:
> > Like we can have another tables shop_logs to store the config versions, while we keep admin_logs just for admin activity.
> 
> I don't think `shop_logs` is any more self-explanatory/self-documenting than `admin_logs` so may as well keep it as-is. And ignoring the config logic then this PR fine by me. 👍

ok - sounds good. I'll keep as is for now then.

backend/db/migrations/20201005202813-addAdminLog.js
franckc: done

Issue #599, opened 2020-10-05
Serve primary shop domain from CDN


nick:
We are almost done with serving the [UI via CDN](https://github.com/OriginProtocol/dshop/issues/495). We should also serve the primary shop from CDN, while ensuring shop data is still pushed to IPFS and the DNS Link is kept active so that a shop can still be requested over IPFS. This will result in better reliability and speed when serving shops.

mikeshultz:
Done for GCP.  Wil drop this to a P2 but leave open for other providers.

mikeshultz:
Not done for GCP and it's looking like that's not going to work as-is.  Bumping back up to a P1.  Google suggests using their load balancers (not going to scale to our current numbers, let alone future), serving using [third party CDNs](https://cloud.google.com/network-connectivity/docs/cdn-interconnect), or [Firebase hosting](https://firebase.google.com/docs/hosting/api-deploy).

Have a request for clarification out to GCP support through a previous quota request but don't expect a response, and we don't have a support contract so that avenue is probably dead in the water.

Going to continue to investigate these options to replace Cloud CDN.

PR #621, opened 2020-10-08
CDN Deployments

mikeshultz:
Deploy shops to a CDN

### TODO

- [x] Deploy to GCP bucket
- [x] Configure balancer
- [x] Configure CloudCDN
- [x] Adjust DNS configuration as necessary
- [x] Make deployments async
- [x] Test Route53
- [ ] Test Cloudflare DNS configuration
- [x] Fix CNAME conflicts when DNS records moved to CDN from IPFS gateway
- [x] Adjust custom domains to work with the CDN deployments

Ref #599


mikeshultz:
I still want to do more testing before merge and deploy, but think it's ready for review whenever you all get the time.

mikeshultz:
Unit tests have not been updated.  Looking at how that might work now.

backend/logic/deploy/bucket/gcp.js
```diff
@@ -0,0 +1,139 @@
...
+
+  return {
+    bucketName: bucket.name,
+    url: `gs://${bucket.name}/`,
+    httpUrl: getHTTPURL(bucket.name)
+  }
+}
+
+/**
+ * Return a configured Storege client
```
franckc, line 86: type: Storage

backend/logic/deploy/bucket/gcp.js
```diff
@@ -0,0 +1,139 @@
...
+ * @param bucket {Object} instance of Bucket
+ * @param updir {string} path to directory to upload
+ */
+async function uploadDirectory(bucket, updir) {
+  const files = await walkDir(updir)
+
+  for (const file of files) {
+    const destination = trimStart(file.replace(updir, ''), '/')
+    log.debug(`Uploading ${file} to gs://${bucket.name}/${destination}...`)
+    await bucket.upload(file, {
```
franckc, line 118: Could we upload files in parallel (maybe in batch so that we control the max number of parallel uploads)? Might be a premature optimization though. I don't have a good crasp of the bucket.upload latency.

backend/logic/deploy/bucket/index.js
```diff
@@ -0,0 +1,58 @@
...
+ *    httpUrl: "https://storage.googleapis.com/mybucket/"
+ * }
+ */
+const { getModules, loadModule } = require('../../../utils/module')
+const { getLogger } = require('../../../utils/logger')
+
+const log = getLogger('logic.deploy.bucket')
+
+/**
+ * Deploy shop to a bucket
```
franckc, line 27: nit: it seems it could upload to multiple buckets if more than 1 module is configured?

backend/logic/deploy/build.js
```diff
@@ -0,0 +1,95 @@
...
+    })
+  })
+
+  let publicShopConfig = {}
+  try {
+    const raw = fs.readFileSync(`${OutputDir}/data/config.json`)
+    publicShopConfig = JSON.parse(raw.toString())
+  } catch (e) {
+    log.warning(`Shop ${shop.id}: failed parsing ${OutputDir}/data/config.json`)
+    // TODO: Under which circumstances would it be ok for this to not be a hard error?
```
franckc, line 41: yeah. I think we should throw. this should never happen

backend/logic/deploy/dns.js
```diff
@@ -0,0 +1,83 @@
+const setCloudflareRecords = require('../../utils/dns/cloudflare')
+const setCloudDNSRecords = require('../../utils/dns/clouddns')
+const setRoute53Records = require('../../utils/dns/route53')
+const { getLogger } = require('../../utils/logger')
+
+const log = getLogger('logic.deploy.dns')
+
+/**
+ * Copy all needed files for a shop deployment to the public build directory and
```
franckc, line 9: This does not seem like the correct description of what the method does? copy/pasta?

backend/logic/deploy/records.js
```diff
@@ -0,0 +1,170 @@
+/**
+ * Funcitonality for deployment records in the database
```
franckc, line 2: typo

backend/logic/deploy/records.js
```diff
@@ -0,0 +1,170 @@
...
+ * Create a shop deployment record
+ *
+ * @param shopId {number} - Shop ID to create a deployment for
+ * @param uuid {string} - A UUID identifier that can be used for later retrieval
+ * @param status {ShopDeploymentStatuses} - ShopDeploymentStatuses enum value
+ */
+async function createDeployment(
+  shopId,
+  uuid,
+  status = ShopDeploymentStatuses.Pending
```
franckc, line 52: why does the status need to be passed? shouldn't a new deployment always get created as Pending? Like we would create the deployment row, then pass it's id to the deploymentProcessor.

backend/logic/deploy/records.js
```diff
@@ -0,0 +1,170 @@
...
+    status: ShopDeploymentStatuses.Success,
+    ...props
+  })
+}
+
+/**
+ * Create a shop deployment record
+ *
+ * @param shopId {number} - Shop ID to create a deployment for
+ * @param uuid {string} - A UUID identifier that can be used for later retrieval
```
franckc, line 46: Why do we need a uuid? Could we simply use shop_deployments.id?

backend/queues/deploymentProcessor.js
```diff
@@ -0,0 +1,54 @@
...
+    networkId: networkId ? networkId : network.networkId,
+    subdomain,
+    shop,
+    pinner,
+    dnsProvider
+  })
+
+  log(90, 'Deployment run completed')
+
+  if (!success && error) {
```
franckc, line 46: Is it possible success would be true but error would be set? under which scenario?

backend/logic/deploy/bucket/gcp.js
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * @param args {Object}
+ * @param args.networkConfig {Object} - Decrypted networkConfig object
+ * @param args.OutputDir {string} - The directory containing the shop build
+ */
+async function deploy({ shop, networkConfig, OutputDir }) {
+  assert(!!networkConfig, 'networkConfig must be provided')
+  assert(!!OutputDir, 'OutputDir must be provided')
+
+  const bucketName = `${BUCKET_PREFIX}${shop.authToken}`
```
franckc, line 56: should we include the networkId to avoid any possible collision between shop deployments on different network but having the same authToken?

franckc (approved):
Looks really good. Super clean code. Love the use of asserts and the great jsdoc.
A few questions inline otherwise LGTM.

We'll definitely want to roll this out with caution since that's a lot of changes. Happy to help testing on staging once this gets merged.

backend/logic/deploy/bucket/gcp.js
mikeshultz: I'm currently not sure if GCP would allow that, so I'd prefer to leave that optimization until later.

backend/logic/deploy/build.js
mikeshultz: This was originally in `utils/deployShop`.  Do you know why this was originally a warning?

mikeshultz: https://github.com/OriginProtocol/dshop/blob/master/backend/utils/deployShop.js#L163

backend/logic/deploy/records.js
mikeshultz: hypothetical future thinking, but really probably unnecessary.  It's doing no harm right now, but you're probably right.

backend/logic/deploy/records.js
mikeshultz: I chose uuid because it makes things a lot cleaner.  A UUID can be created before the deployment process starts without having to fiddle with deployment locking DB records.

Specifically, the deployment DB record is first created in `deploy()` function, which is called in a bull queue job.  The only alternative would be to create the deployment DB record ahead of time, which doesn't really make sense, IMO.  Leaves complexity in the job processor instead of the express handler.

backend/queues/deploymentProcessor.js
mikeshultz: Not currently, no

backend/logic/deploy/bucket/gcp.js
mikeshultz: I hadn't considered that.  Do we allow any `authToken` collisions in an instance?  I don't think so.  But since we may have multiple instances(for us, staging/prod), `BUCKET_PREFIX` should probably at least reflect that.

mikeshultz: I added `NODE_ENV` into the prefixes.

backend/logic/deploy/build.js
franckc: yeah. the code was eating that exception but that did not make any sense to me so I added the warning + TODO.
can you maybe check the warning does not show up in the logs? And if not let's throw an exception.

backend/logic/deploy/records.js
franckc: Sounds good. I'll leave it up to you to decide.

backend/logic/deploy/bucket/gcp.js
franckc: We don't allow collision on an instance (we check the authToken does not already exist in the DB before using it for a new shop).

I believe NODE_ENV is set to "production" for both staging and prod. I'd be slightly inclined to use networkId.

mikeshultz: > I believe NODE_ENV is set to "production" for both staging and prod so I'd be inclined to use networkId.

Well, that makes this tricky.  I still think there's a good possibility in the future for an individual node to support multiple networks, so I'm hesitant to use networkId.  Though... we're depending on networkConfig for credentials right now.  Maybe that is a good choice.

backend/logic/deploy/bucket/gcp.js
nick: I found a [code sample](https://github.com/googleapis/nodejs-storage/blob/master/samples/uploadDirectory.js#L69) that uses `Promise.all` to do it... but can optimize later

backend/logic/deploy/build.js
mikeshultz: There doesn't appear to have been any occurrences of this in the logs, at least within the last 6 months.  Will remove the catch here.

mikeshultz: Actually, just going to throw a more descriptive error.

nick (approved):
I read through the code and it looks good. Will try to run it locally too...

mikeshultz:
Added some items to the TODO

mikeshultz:
updated tests in https://github.com/OriginProtocol/dshop/pull/621/commits/ef82940137a27bedc945550c99bf2ee381254846.  Not super excited about the implementation of function overrides but it'll do.  Feedback welcome.

Issue #628, opened 2020-10-12
New products added disappears after publishing shop


gullible1:
I tried multiple times to add new products to my store [annas-cabanas.ogn.app](url) but each time after clicking "publish changes", they would disappear from my list of products. 

https://www.loom.com/share/f9a804576cc743008e8317c44773793b

PR #629, opened 2020-10-12
Remove homeCollection filter

shahthepro:
Removing this since it causes #628 

Alternatively, we can use `useProducts({ collection: 'home' })` to achieve the same thing without causing the above bug. Once we are done with themes' editor, we could define featured collections/products as a field.


shahthepro:
Merging and deploying this now

PR #648, opened 2020-10-19
Fix #636

### LINKED ISSUE #636: Stripe integration/connection keeps getting lost
The owners of the [Argent store](https://dshop.originprotocol.com/argent.ogn.app) has reported that their Stripe integration/connection keeps getting lost. It would be successfully connected and then after a while, become disconnected again. If the merchant saves, then refreshes the page, the Stripe connection disappears every time.

shahthepro:
Avoids disconnecting Stripe when other config are being updated.


franckc (commented):
Great catch! Thanks for fixing this

PR #649, opened 2020-10-19
Uni theme

nick:


shop/src/components/SocialLinks.js
```diff
@@ -1,47 +1,74 @@
...
+const SocialLinks = ({
+  el = 'div',
+  className = 'social',
+  itemClassName,
+  svg = {}
+}) => {
   const { config } = useConfig()
-  const social = config.twitter || config.medium || config.instagram
-  if (!social) {
+  const social = pick(config, 'twitter', 'medium', 'instagram', 'youtube')
```
shahthepro, line 19: @nick Shouldn't the second arg be an array (instead of spread-out array)?

Issue #379, opened 2020-08-05
Shop health check endpoint


nick:
I'm thinking we should have a basic health check endpoint that can be polled on occasion to check:

* All API keys are functioning
* All relevant webhooks are correctly configured
* All frontend configs (eg config.json) adhere to schemas
* External payments have corresponding orders
* Other things... (please comment)

We can also show the results of the health check in the admin

mikeshultz:
A shop diagnostics endpoint.  :+1: 

Related but probably not relevant: there's a couple of health checks for the backend itself under `/health`.

franckc:
+1 Good idea!

franckc:
Could other data that might be helpful:
 - any job in the makeOffer or event queue that failed and that are associated with the shop
 - deployment failures

The more tools we provide to the merchants for self-diagnosing issues, the less support work we'll have to handle.

franckc:
Implemented in #663

PR #663, opened 2020-10-22
Implement a new shop config diagnostic endpoint

franckc:
Part of the work on #379 

This PR implements a new endpoint /health/shop that performs a bunch of tests on the shop configuration.
In a separate PR I'll add a basic page in the shop's admin tool to call the back-end and display the data.

Note: there are additional checks that could be added (ex: test paypal webhook, prinful credentials, etc...) but in the spirit of getting something out quickly and iterating, I'll leave those out of this PR for now.


backend/logic/shop/health.js
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  //
+  // If the shop is published, load config.json from the web and check its validity.
+  //
+  errors = []
+  let shopJsonConfigOnWeb = {}
+  if (shopIsPublished) {
+    try {
+      const url = `shopConfig.dataUrl/config.json`
```
mikeshultz, line 216: missing the wrapper(`${shopConfig.dataUrl}`) here?

mikeshultz (changes requested):
Some commentary but think I did find one string formatting bug which would fail the config.json checks.

Looks good overall!  :+1:

backend/logic/payments/stripe.js
```diff
@@ -1,13 +1,142 @@
...
+ *
+ * @param {models.Network} network
+ * @param {models.Shop} shop
+ * @returns {Promise<{success: false, error: string}|{success: true}>}
+ */
+async function checkStripeConfig(network, shop) {
+  const networkConfig = getConfig(network.config)
+  const validWebhhokUrls = [
+    `${networkConfig.backendUrl}/webhook`,
+    `https://dshopapi.ogn.app/webhook` // Legacy URL. Still supported.
```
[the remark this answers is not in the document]
nick: Yea I think we should flag old URLs as bad so we can fix them up

backend/logic/payments/stripe.js
```diff
@@ -1,13 +1,142 @@
...
+  if (!stripeKey) {
+    log.info('No stripe key configured')
+    return
+  }
+  const stripe = Stripe(stripeKey)
+
+  // Load the last 100 shop's orders paid with Stripe and get their encrypted IPFS hashes.
+  const orders = await Order.findAll({
+    where: { shopId, paymentType: OrderPaymentTypes.Stripe },
+    limit: 100,
```
nick, line 82: The doc above says every order in the last 30 days... should we query based on createdAt and remove the limit? Also perhaps limit attributes to `encryptedIpfsHash` since that's the only field we need? `attributes: ['encryptedIpfsHash']`

backend/logic/shop/health.js
```diff
@@ -0,0 +1,332 @@
...
+      const res = await fetch(url)
+      shopJsonConfigOnWeb = await res.json()
+      errors = checkJsonConfig(
+        shop.id,
+        network.networkId,
+        shopJsonConfigOnWeb,
+        'Json config on web'
+      )
+    } catch (e) {
+      log.error(`Shop {shop.id} - Failed loading json config on web: ${e}`)
```
nick, line 226: `Shop ${shop.id}`

nick (approved):

backend/logic/payments/stripe.js
franckc: Done.

backend/logic/shop/health.js
franckc: oops! good catch

backend/logic/shop/health.js
franckc: Fixed. Ty.

franckc:
I ended up including a simple UI that shows the shop diagnostic data under the admin Console page.
Merging this in now.

PR #665, opened 2020-10-22
Switch to static classNames and bug fixes

shahthepro:


shahthepro:
Merging and deploying this now, let me know if you have any comments, will address it later

PR #668, opened 2020-10-22
Node 14 upgrade

nick:


mikeshultz:
I sort of remember looking at this not too long ago and running into an issue but I can't remember what it was (assuming a real memory).  I went looking but everything checks out and looks good.  

Have you build-tested the backend with 14?   Any API changes we have to worry about?  We cool with diverging from the `origin` repo?  

Also, I would like to wait for #621 to settle before rolling this if there's no urgency here.

nick:
Yep no urgency.

I haven't build tested it yet, no. Don't think there should be any issues though

mikeshultz (approved):
Read through the release notes for v11-14 and nothing jumped out at me.  Did some local tests, including a docker build.  Everything looks good.

I took care of the merge conflicts.

Maybe wait until Monday to deploy but LGTM

Issue #669, opened 2020-10-22
Improve CDN deployment UX


mikeshultz:
Want to track any UX issues with CDN deployments here.

### SSL Certificate Provisioning

Right now, when creating a new CDN endpoint, [GCP can take up to 60 minutes](https://cloud.google.com/load-balancing/docs/ssl-certificates/troubleshooting?&_ga=2.165840060.-360165820.1597679396#certificate-managed-status) to issue and deploy the cert.

In practice, I have yet to see it take that long(usually minutes) but it may still be confusing to users.  In the mean time, before provisioning, Chrome returns the `ERR_SSL_VERSION_OR_CIPHER_MISMATCH` error.

This gets a little more confusing when switching from a CNAMEd IPFS deployed shop to a CDN.

I think the deployment module should probably wait on this until a successful SSL connection can be made.  Though I'm not sure if a 5-15minute spinner is really ideal here.

### Cache confusion

I think we might want to play this by ear to begin with.  The CDN bucket backend uses a type of caching that may not show updates immediately.  So far in my testing, I haven't noticed the cache last an unreasonable amount, but a user may expect immediate updates and not see them.

If this ends up being a pain point, we should consider cache invalidations on deployment.  However, cache invalidations are also fairly slow(tests show ~5min for GCP to say it's complete, but functionally ~10min).

**NOTE: Cache invalidation has been implemented**

### Initial DNS switch

The switch from the IPFS CNAME to the CDN will take time to propagate.  This could lead to confusion, but I think is generally a low priority.  It's also a one-time affair.  If anyone thinks we should pay more attention to this before rolling prod, lmk.

### [DONE] Domain not showing on the admin > Settings > Publish page
Legacy publishes are missing the domain on that page.

**NOTE:** `shop_domains` table backfilled

mikeshultz:
Split this up into individual issues.  #683 #684 #685

PR #671, opened 2020-10-22
Adds urlMaps cache invalidation

mikeshultz:
Adds cache invalidations during deployments if a urlMaps is not newly created.

Especially check out the invalidation paths and let me know if I'm missing anything crucial.

Ref: #669

### DIFF
--- a/backend/logic/deploy/cdn/gcp.js
+++ b/backend/logic/deploy/cdn/gcp.js
@@ -124,7 +124,14 @@ async function configureCDN({ shop, deployment, domains }) {
    * Ours is basically wildcards direct to the bucket backend.
    */
   let urlMap = await getUrlMap(rest, serviceName)
-  if (!urlMap) {
+  if (urlMap) {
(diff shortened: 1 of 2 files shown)


backend/logic/deploy/cdn/gcp.js
```diff
@@ -281,6 +288,29 @@ async function createUrlMap(client, backendBucketLink, name) {
...
+/**
+ * Invalidate a urlMap cache
+ *
+ * @param client {object} REST client with Oauth
+ * @param urlMapName {string} name (resource ID) of the urlMap
+ * @param paths {Array<string>} paths to invalidate (starting with /)
+ */
+async function invalidateCache(client, urlMapName, paths) {
+  for (const pth of paths) {
+    await post({
```
franckc, line 300: Maybe add a comment that we could optimize this in the future by making concurrent requests.

franckc (approved):
LGTM

PR #676, opened 2020-10-23
Reconfigured CDN when a domain is added

mikeshultz:
What should be straight forward turns convoluted.  This updates the CDN when a domain is added by reconfiguring the CDN.  The CDN module(s) will notice the change in domains and adjust resources as necessary.  In the GCP case, create a new cert, create a new proxy, and swap in the new proxy.

I'm going to merge this just in case a prod deploy happens but post-merge reviews welcome.


backend/routes/domains.js
```diff
@@ -31,6 +33,32 @@ module.exports = function (router) {
...
         status: ShopDomainStatuses.Pending
       })
+
+      // Re-configure CDN to pick up domain changes
+      const deployments = await ShopDeployment.findAll({
+        where: {
+          shopId: req.shop.id
+        },
+        limit: 1,
+        order: [['createdAt', 'DESC']]
```
franckc, line 43: nit: order by id is more performant since there is no DB index on createdAt

mikeshultz: IDs technically aren't guaranteed to be sequential and is generally considered bad form.

Also, index performance boosts to sorting [are kind of minimal for large datasets and sometimes negative, relying more on disk IO instead of memory/CPU, causing disk thrashing on standard disks](https://www.postgresql.org/docs/9.3/indexes-ordering.html).  And considering the size of the dataset here, I think we're fine.

franckc: Agreed data set is small so we don't need to worry.

Regarding "IDs technically aren't guaranteed to be sequential and is generally considered bad form", ids are generated by a sequence and therefore are guaranteed to be monotonically increasing, which is what we care about for ordering I think.

mikeshultz: Sequences can wrap.  INSERTs can be done with explicit IDs skipping sequences entirely.  While these conditions may be unlikely for our dataset, they're still considered bad form because of that.  ID ordering is implicit, while the createdAt date is explicit.

PR #678, opened 2020-10-23
Printful pagination and description override

### LINKED ISSUE #601: Allow seller to edit Printful product descriptions
Currently sellers have no choice but to use the product descriptions provided by Printful. We should let them have the option of editing the description so that they can add more detail about their designs.

shahthepro:
- [x] Allow users to have different description of printful products (Closes #601)
- [x] Paginate during Printful sync


nick (approved):

franckc (commented):
LGTM

PR #705, opened 2020-10-27
Reset all config fields when Stripe/Paypal/Printful is disconnected

franckc:
We have some instances of shops in production that have inconsistent configs in the DB where for example a webhook is not registered but the shop does seem to be configured for Printful.

This PR explicitly resets all the config fields related to Stripe/PayPal/Printful whenever one of those app is disconnected by the admin. This should hopefully help with inconsistency issues moving forward. I'll write a script in a separate PR to fix the existing data inconsistencies.


shahthepro (approved):
LGTM.

nick (approved):

PR #707, opened 2020-10-27
[WIP] Refactor Printful logic and add a script to check printful webhook configuration

franckc:


nick:
I looked through and it looks good, but wondering if we check for non-Dshop webhooks to make sure we don't overwrite existing webhooks for other integrations (eg Shopify). I think Printful only allow one webhook per store and don't think we should overwrite existing ones without first asking the merchant

nick (approved):

franckc:
@nick Thanks for double checking. Yes I have a check in place to not de-register the webhook in case it points at a non-dshop URL. See https://github.com/OriginProtocol/dshop/pull/707/files#diff-2228f193780c8b9717c04786a4c590415bb3018659050262db90710b607d84d9R128 and https://github.com/OriginProtocol/dshop/pull/707/files#diff-3b923a18d23661d7c4737d0167542ff359606ec9670cccf500f86e4b43a101d8R270

PR #709, opened 2020-10-27
Disable CDN deployments

mikeshultz:
Not ready to deploy this just yet(still looking for alternatives), and hopefully this is only temporary but I want it ready to go before we hit the next quota.

That said, it's looking increasingly likely that Cloud CDN deployments aren't going to work out near-term so this PR disables them.  Already deployed shops will continue to work fine and move back over the the IPFS gateway upon next deployment.

Bucket deployments will continue to happen, and any custom domains already pointing at a GCP balancer will continue to work, until we decide its necessary to clean up the Cloud CDN/load balancer resources.


franckc (approved):
LGTM in case we need to use it.

mikeshultz:
Replaced with #713

PR #713, opened 2020-10-28
Optional CDN deployments per shop

mikeshultz:
Adds per-shop CDN opt-in, currently to be enabled via direct DB access.

I'm not sure where the UI would this should lay so I'm leaving that for later.  Perhaps in super-admin?  I imagine this decision should be by whoever is providing the infra.

This replaces #709


franckc (approved):
LGTM

PR #722, opened 2020-10-29
Add retries to GCS bucket uploads

mikeshultz:
n/t

### DIFF
--- a/backend/logic/deploy/bucket/gcp.js
+++ b/backend/logic/deploy/bucket/gcp.js
@@ -129,6 +129,30 @@ async function getBucket(bucketName) {
 }
 
 /**
+ * Upload a file to destination in bucket
+ *
+ * @param bucket {object} Bucket instance
+ * @param file {string} full path to filename
+ * @param destination {string} destination name in bucket
+ * @param attempt {number} Number of attempts that have so far (internal)
+ * @returns {object} UploadResponse object
+ */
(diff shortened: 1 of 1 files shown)


nick (approved):

backend/logic/deploy/bucket/gcp.js
```diff
@@ -128,6 +128,30 @@ async function getBucket(bucketName) {
...
+ */
+async function uploadFile(bucket, file, destination, attempt = 0) {
+  try {
+    return await bucket.upload(file, {
+      destination,
+      public: true
+    })
+  } catch (err) {
+    if (err && err.code === 502 && attempt < 3) {
+      return await uploadFile(bucket, file, destination, attempt + 1)
```
franckc, line 148: haha. neat. the use of recursion will make the stack trace longer but no big deal.

franckc (approved):
LGTM

backend/logic/deploy/bucket/gcp.js
franckc: also do we think an immediate retry without a delay will succeed? I guess let's see...

mikeshultz: I have no reason to suspect a delay would benefit this at all.  AFAIK, it's not a rate limit.  If we see otherwise, we can change it.

PR #725, opened 2020-10-30
Cleanup disconnect logic

franckc:
I introduced a bug where Printful would get disconnected upon a config update :(

This should hopefully fix the issue and clean up our disconnect logic which was a bit confusing for a newbie like myself.
The main change is for the front-end disconnect modal to explicitly specify if an app should be disconnected as opposed to relying on the back-end figuring it out if the config change is related to connecting or disconnecting an app.


franckc:
Merging this in so I can deploy a fix to prod asap.

Happy to address any comment in a separate PR...

PR #730, opened 2020-10-30
Add Inventory management

shahthepro:
Pending things:
- [x] Add tests
- [x] Increment stocks on order cancelation and rejection


backend/db/migrations/20201029063356-addProducts.js
```diff
@@ -0,0 +1,29 @@
...
+
+    return queryInterface.sequelize.transaction(async () => {
+      await queryInterface.createTable(tableName, {
+        id: {
+          type: Sequelize.INTEGER,
+          autoIncrement: true,
+          primaryKey: true
+        },
+        product_id: Sequelize.STRING,
+        stock_left: Sequelize.INTEGER,
```
franckc, line 15: Should this be an unsigned integer since it should never go negative?

backend/logic/order/index.js
```diff
@@ -194,6 +194,30 @@ async function processNewOrder({
   const order = await Order.create(orderObj)
   log.info(`Saved order ${order.fqId} to DB.`)
 
+  if (shopConfig.inventory) {
+    const cartItems = get(data, 'items', [])
+    const dbProducts = await Product.findOne({
+      shopId: shop.id,
+      productId: cartItems.map((item) => item.product)
```
franckc, line 201: I did not realize findOne can take a list of productIds... and there is not need for a where clause?

backend/logic/order/index.js
```diff
@@ -194,6 +194,30 @@ async function processNewOrder({
...
+    })
+
+    for (const product of dbProducts) {
+      const item = cartItems.find((item) => item.product === product.productId)
+      const variantId = get(item, 'variant')
+
+      if (variantId) {
+        product.update({
+          stockLeft: product.stockLeft - item.quantity,
```
franckc, line 210: What if there is not enough stock? The Front-end does some checks but in case of multiple buyers sending their order in a short timespan, shouldn't the back-end check?

It may not be easy to handle with our current asynchronous architecture...
I'm actually a similar problem while looking at the issue related to over-use of a discount code on the Injective shop. The back-end did not refuse those orders where the discount was over-used. 

But let's think about possible solutions...

backend/logic/order/index.js
```diff
@@ -194,6 +194,30 @@ async function processNewOrder({
   const order = await Order.create(orderObj)
   log.info(`Saved order ${order.fqId} to DB.`)
 
+  if (shopConfig.inventory) {
+    const cartItems = get(data, 'items', [])
+    const dbProducts = await Product.findOne({
```
franckc, line 199: What if any of the item in the cart is not found in the DB? shouldn't this be an error?

backend/logic/shop/config.js
```diff
@@ -40,6 +40,7 @@ const dbConfigOptionalKeys = [
   'deliveryApi',
   'email', // Email provider: 'aws', 'sendgrid', 'mailgun'
   'emailSubject',
+  'inventory',
```
franckc, line 43: Add a comment explaining this is a boolean indicating if product inventory is enabled/disabled for the shop

shop/src/data/fetchProductStock.js
```diff
@@ -0,0 +1,19 @@
+import memoize from 'lodash/memoize'
+
+const fetchProductStock = memoize(
```
franckc, line 3: Are we sure we want to cache this value given the inventory will change as buyers get the item?

backend/logic/products/index.js
```diff
@@ -358,6 +399,13 @@ async function deleteProduct(shop, productId) {
   writeProductsFile(shop, updatedProducts)
   await removeProductData(shop, productId)
 
+  await Product.destroy({
```
franckc, line 402: What happens if there is a pending order for the item (ex: due to the use of a PayPal echeck) - will that cause any issue processing the order?

Rather than deleting the row, an alternative would be to have a status column with enum values Active/Inactive.

franckc (commented):
Took a quick pass. Great clean up of the code! :)
Overall looks good but a few questions/comments inline.

backend/db/migrations/20201029063356-addProducts.js
shahthepro: Yep, will change it

backend/logic/order/index.js
shahthepro: Sorry, I'm might have messed up this, will fix this

shop/src/data/fetchProductStock.js
shahthepro: Without caching, there are too many requests to fetch the stock. That's the reason I added it.

backend/logic/products/index.js
shahthepro: Gotcha, will add it

shahthepro: Have more questions regarding this, even if we add the status column, a user could still buy an "Inactive" item. Also, the productId conflicts (in a shopId) could cause some weird things since `Product.findOne` is used.

backend/logic/order/index.js
shahthepro: I added some checks for it but it justs add an error message to the orderObj instead of throwing. Because if we throw and don't create the Order, there will be no track of the order in Dshop and sellers have to search through their PayPal or Stripe accounts to process the refund

backend/db/migrations/20201029063356-addProducts.js
shahthepro: There seems to no unsigned int in postgres. But have added a less-than-zero check before we update the count. That should be fine for now I think

shahthepro:
@franckc Have added some tests and addressed most of your comments other than the `status` column

franckc (approved):
A few comments inline otherwise LGTM.

One thing we should discuss (and that could be implemented in a separate PR) is a pre-flight check right before the user attempts to pay. This would be to lower the chance that when multiple buyers are trying to buy a product with let's say only 1 stock left at the same time, they currently can all complete the checkout flow and pay for the product. Even if the product is actually depleted. And we don't let them know until the merchant goes under admin and refunds the order. This is a pretty poor buyer experience.
The idea of a pre-flight would be that right when the user clicks "Pay" we send a request to the back-end to check on the product availability. There is still a chance of a race condition between the pre-flight the payment being processed then the order created. But it would reduce it hopefully.

backend/logic/order/index.js
```diff
@@ -190,6 +190,21 @@ async function processNewOrder({
     orderObj.data.error = error
   }
 
+  // Note: Not throwing here because it could
+  // cause if the order isn't created on Dshop
```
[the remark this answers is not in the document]
shahthepro: Yeah, will change it

backend/logic/order/index.js
```diff
@@ -190,6 +190,21 @@ async function processNewOrder({
...
+  // and other payment methods
+  const { error: inventoryError } = await updateInventoryData(
+    shop,
+    shopConfig,
+    data
+  )
+
+  orderObj.data.inventoryError = inventoryError
+  if (!orderObj.data.error) {
+    orderObj.data.error = inventoryError
```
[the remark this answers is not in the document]
shahthepro: I put this inside the `!orderObj.data.error` so that it doesn't overwrite. But yeah, an array of errors sound good too

backend/logic/order/index.js
```diff
@@ -373,6 +410,106 @@ async function processDeferredPayment(paymentCode, paymentType, shop, event) {
...
+  })
+
+  const allValidProducts = cartItems.every(
+    (item) => !!dbProducts.find((product) => product.productId === item.product)
+  )
+
+  if (!allValidProducts) {
+    log.error(`[Shop ${shop.id}] Invalid product ID`, cartItems)
+    return {
+      error: 'Some products in your cart are unavailable'
```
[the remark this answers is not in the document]
shahthepro: Will do it

shop/src/data/fetchProductStock.js
shahthepro: Leaving the `productId` undefined will do the same thing.

PR #731, opened 2020-10-30
Serve dapp from backend if given a deployed hostname

mikeshultz:
This PR allows us to serve assembled shops from the backend by pointing a known domain at the backend.

This will require some modifications to the nginx ingress to set the `X-Forwarded-Host` header for prod deployments.

Will follow this up later with an update to DNS instructions stuff but want to get in some testing before that happens.

### DIFF
--- a/backend/utils/hostCache.js
+++ b/backend/utils/hostCache.js
@@ -1,0 +1,57 @@
+/**
+ * Singleton to keep track of hostname->auth_token for speedy lookups for static
+ * assets.
+ */
+
+const { ShopDomain } = require('../models')
+const { getLogger } = require('../utils/logger')
+
+const log = getLogger('utils.hostCache')
+let lastLookupCleanup = 0
+let lastLookup = {}
+const hostnames = {}
+
+const LOOKUP_LIMIT = 3000 // 3 seconds
+const LOOKUP_CLEANUP_INTERVAL = 900000 // 15 minutes
+
+/**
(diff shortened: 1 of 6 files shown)


backend/utils/hostCache.js
```diff
@@ -0,0 +1,57 @@
...
+    lastLookupCleanup = now
+  }
+
+  if (typeof hostnames[host] === 'undefined') {
+    log.debug('Cache miss')
+
+    // We don't want to poke the DB too often for static file requests
+    if (!lastLookup[host] || now - lastLookup[host] > LOOKUP_LIMIT) {
+      const shopDom = await ShopDomain.findOne({
+        where: { domain: host },
```
franckc, line 40: data is tiny for now so performance will be fine. but might as well add an index on the 'domain' column for the future.

franckc (approved):
LGTM

Do we have any concern about the performance of serving those files from Google NFS disk? I guess if a shop starts driving decent traffic volume we'll upgrade them to CDN so probably not an issue.

mikeshultz:
> Do we have any concern about the performance of serving those files from Google NFS disk?

NFS reads are pretty good.  There's caching at the OS level.  And as you say, if they get heavy they can be moved to the CDN.

backend/utils/hostCache.js
mikeshultz: Great point, I'll add it to this PR.

PR #735, opened 2020-11-02
Do not allow an order amount should to go negative when a discount is used

franckc:
- Apply max(totalAmount, 0) to prevent an order amount to ever go negative if a fixed discount that is larger than the total is applied
 - Refactored some of the discount code and moved it from backend/utils to backend/discount
 - Took out the option on the front-end to not apply discount on shipping since this was never implemented on the backend
 - Added some unit tests

Left TODO (will handle in a separate PR in order to keep things easier to review):
 - On the front-end, if the total amount is zero after applying a discount, the checkout page does not let the user submit
 - Add a pre-fight on checkout to verify the availability and validity of a discount


backend/logic/discount/index.js
```diff
@@ -195,25 +172,25 @@ const validateDiscountOnOrder = async (
 
   const donation = get(cart, 'donation', 0)
 
-  const calculatedTotal = subTotal + shipping - discount + donation + totalTaxes
+  // Note: By calling Math.max, we don't let the amount go negative in case a
+  // discount larger than the total is applied.
+  const calculatedTotal = Math.max(
+    0,
+    subTotal + shipping - discount + donation + totalTaxes
```
mikeshultz, line 179: Know this was touched upon in discord this morning, but do we really want to apply discount after shipping?  Shipping is generally a fixed cost which could end up costing the seller.  

If it's fine with everyone, it's fine with me, just through I'd confirm.

mikeshultz (approved):
LGTM, one consideration inline

backend/logic/discount/index.js
franckc: I created this separate issue to track this: https://github.com/OriginProtocol/dshop/issues/745

PR #737, opened 2020-11-03
DNS instructions to use backend address

mikeshultz:
Changes DNS instructions so that instead of pointing to the IPFS gateway assuming it has AutoSSL configured, to use:

- If CNAME: the hostname from `networkConfig.backendUrl`
- if A: `EXTERNAL_IP` if configured, or the external IP as seen by an ipify request, or 127.0.0.1 if localhost

Depends on: https://github.com/OriginProtocol/origin/pull/4510

### DIFF
--- a/backend/utils/const.js
+++ b/backend/utils/const.js
@@ -67,7 +67,8 @@ const {
   REDIS_URL,
   IPFS_GATEWAY, // IPFS gateway override
   BUCKET_PREFIX = DEFAULT_BUCKET_PREFIX,
-  SERVICE_PREFIX = DEFAULT_SERVICE_PREFIX
+  SERVICE_PREFIX = DEFAULT_SERVICE_PREFIX,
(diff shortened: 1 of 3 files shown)


backend/utils/const.js
```diff
@@ -78,6 +79,9 @@ const {
 const DATA_URL = null
 const PRINTFUL_URL = 'https://api.printful.com'
 
+// Service that returns a plaintext IP for a GET request
```
franckc, line 82: Service is free and does not have any rate limiting we need to worry about?

franckc (approved):
LGTM. Just one question inline.

backend/utils/const.js
mikeshultz: https://www.ipify.org/

Though, infinite-caching this would probably be a good idea, since it shouldn't change during an instance run.

PR #742, opened 2020-11-03
CNAME backend host for defualt domains

mikeshultz:
This will by default, CNAME shops on the default domain (e.g. `myshoop.ogn.app`) to the hostname from `backendUrl`.


franckc (approved):
LGTM

PR #744, opened 2020-11-04
Replace encConfig.get with decryptConfig

### LINKED ISSUE #740: Shop pulling shipping rates from Printful but showing free shipping
Orders were placed on the Argent store, which is pulling shipping rates from Printful. However the orders are showing "free shipping" even though it's not free shipping on Printful, and the shipping fee isn't passed on to the buyer.

shahthepro:
Also fixes #740


franckc (approved):
Thanks so much for cleaning this up! LGTM

Let's deploy to prod so that we don't see these weird config issues anymore.

PR #749, opened 2020-11-05
Infra piecemeal config and AWS support

mikeshultz:
This PR has a little bit of noise.  Here's a high level summary of what was done:

- Refactors deployment for explicit infra resource selection instead of implicit from configured credentials.  For now, it relies on network-level default selection, but in the future may have each shop/deployment use a different configuration.
- Creates `@origin/dshop-validation` which currently holds infrastructure settings validations used by both the dapp and backend.  It's primary config is done via JSON which describes dependencies and required configuration needed for each infra resource to work.
- Adds network level piecemeal infrastructure configuration
- Adds super-admin UI for infra resource selection for each network
- Adds a couple backend API endpoints to provide available resources to client
- Combines enums files because it was confusing
- Adds AWS S3 bucket support

I still need to add Cloudfront support but figure this PR has grown enough.  Will follow up with addition of Cloudfront support.


backend/utils/const.js
```diff
@@ -90,6 +90,14 @@ const THEMES_CACHE = IS_TEST
   ? TEST_THEMES_CACHE
   : process.env.THEMES_CACHE || `${__dirname}/../themes`
 
+const DEFAULT_INFRA_RESOURCES = [
+  'gcp-files',
+  'gcp-dns',
+  'ipfs-pinata',
```
franckc, line 96: super super nit: to follow the standard vendor-service used by the other entries, this should be named 'pinata-ipfs'

franckc (approved):
Looks great to me. 

Thanks for the attention you pay to having jsdoc and comments throughout the code. And for adding those unit tests.

Nice work sir!

It's huge PR, so when you merge it to master, let's test it on staging before we deploy to prod.

backend/utils/const.js
mikeshultz: This kind of bugs me too.  But if I change it then I'd have to do `ipfs-cluster-ipfs` which is ridiculous!  Or arguably it could be `ipfs-pinata-files` and `ipfs-cluster-files` but the deployment process between buckets and IPFS are separate enough that I don't want to do that either.  Maybe I should just give them all numbers and be done with it!

I could honestly hand-wring over this for days.

franckc: haha. yeah, don't worry about. not worth overthinking this. lgtm as is.

PR #746, opened 2020-11-05
Payment-specific discounts

shahthepro:
- [x] Add payment-specific discounts
- [x] Add `minCartValue` and `maxDiscountValues`

PR #751, opened 2020-11-06
Fix other discounts

shahthepro:
Fix non-payment-specific discounts that seems to be broken because of my previous PR #746

### DIFF
--- a/backend/logic/discount/index.js
+++ b/backend/logic/discount/index.js
@@ -11,27 +11,32 @@ const { Sequelize, Discount, Order } = require('../../models')
  * @returns {{error: string}|{discount: models.Discount}}
  */
 const validateDiscount = async (code, shop) => {
-  const discounts = await Discount.findAll({
-    where: {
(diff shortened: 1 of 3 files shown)


backend/logic/discount/index.js
```diff
@@ -11,27 +11,32 @@ const { Sequelize, Discount, Order } = require('../../models')
...
+    },
+    endTime: {
+      [Sequelize.Op.or]: [
+        { [Sequelize.Op.gt]: Date.now() },
+        { [Sequelize.Op.eq]: null }
+      ]
     }
+  }
+
+  if (!code) {
```
franckc, line 34: Why would the validateDiscount method ever be called without a code?
It's not obvious to me. Can you add a comment in the code explaining that so that we don't scratch our head in the future when reading this code?

shahthepro: Payment-specific discounts don't have a code. Will add a comment.

PR #754, opened 2020-11-06
Fix `minCartValue` and add some comments

shahthepro:


backend/logic/discount/index.js
```diff
@@ -32,6 +32,7 @@ const validateDiscount = async (code, shop) => {
   }
 
   if (!code) {
+    // Discounts of type 'payment' won't have a code
```
franckc, line 35: Wonder if it would be worth assigning an internal code (like 'PAYMENT') at creation tie to those automatic discounts as opposed to leaving that column empty? Might make it easier when we look at the data. Not a must-have. Just a thought.

backend/models/discount.js
```diff
@@ -22,7 +22,7 @@ module.exports = (sequelize, DataTypes) => {
       discountType: {
         type: DataTypes.ENUM(DiscountTypeEnums)
       },
-      // If discountType is 'fixed', value is in dollar (not cents). Ex: 30 -> $30
+      // If discountType is 'fixed' or 'payment', value is in dollar (not cents). Ex: 30 -> $30
```
franckc, line 25: I'm confused. I thought the automatic discounts were a percentage?

shop/src/data/state.js
```diff
@@ -352,20 +352,16 @@ const reducer = (state, action) => {
...
-    'cart.discountObj',
-    {}
-  )
+  const { minCartValue, maxDiscountValue, discountType } = discountObj
 
   let discount = 0
 
-  if (!minCartValue || newState.cart.subTotal > minCartValue) {
+  const totalWithShipping = get(newState, 'cart.subTotal', 0) + shipping
+  if (!minCartValue || totalWithShipping > minCartValue * 100) {
```
franckc, line 360: Maybe we should move all the cart calculation code that is shared between backend and front-end to packages/utils/cart.js so that we don't get into a situation where the back-end cart calc does not match the ones from the front-end.

Not a must have in this PR we can add a TODO / handle in a separate PR.

backend/logic/discount/index.js
```diff
@@ -199,8 +196,8 @@ const validateDiscountOnOrder = async (orderObj) => {
       }
     }
 
-    if (maxDiscountValue) {
-      discount = Math.min(maxDiscountValue, discount)
+    if (maxDiscountValue && ['percentage', 'payment'].includes(discountType)) {
```
franckc, line 199: How does maxDiscountValues get set for an Automatic discount? I don't see that option being available in the admin UI? I assume it's for the future (which is a good thing to have)?

franckc (approved):
A few comments/questions inline otherwise lgtm

backend/logic/discount/index.js
shahthepro: We should also add a `discountType != 'payment'` constraint to where clause if we end up doing this. 
Also, it will make the code `PAYMENT` a reserved one

backend/models/discount.js
shahthepro: Oh, yes, mistake on my end, will change this comment

shop/src/data/state.js
shahthepro: Yep, agreed

backend/logic/discount/index.js
shahthepro: Yeah, at present, you can set the `maxDiscountValue` for only 'percentage' discountType. The UI doesn't have a field for this in automatic discounts screen.

PR #758, opened 2020-11-09
Bug fixes

shahthepro:


backend/test/utils.js
```diff
@@ -351,13 +351,7 @@ async function createTestEncryptedOfferData(
...
       label: 'Credit Card'
     },
-    discountObj: discount
-      ? {
-          code: discount.code,
-          discountType: discount.discountType,
-          value: discount.value
-        }
-      : {},
+    discountObj: discount ? discount : {},
```
franckc, line 354: Isn't that going to break the tests that don't pass a discount?

franckc (commented):
Great refactor. thank you! one comment below regarding test/utils.js

backend/package.json
```diff
@@ -7,6 +7,7 @@
   },
   "main": "index.js",
   "scripts": {
+    "start": "nodemon index.js",
```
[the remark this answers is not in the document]
shahthepro: Sure, will add it to readme

backend/test/utils.js
shahthepro: No, it shouldn't. It just passes all the fields of discountObj instead of picking only 3.

franckc: yes but I was wondering in case no discount is passed as an argument to this method. then discountObj would be set to undefined. which may cause problem?
anyway, I assume you verified all the tests ran so I'm likely wrong.

PR #772, opened 2020-11-11
Add and populate an active column on etl_shops

franckc:


backend/etl/processor.js
```diff
@@ -130,10 +132,62 @@ class EtlJobProcessor {
...
-    // TODO(franck): figure out how to extract this information.
-    log.warn(`Shop ${shop.id}: custom domain extraction not implemented yet.`)
-    return null
+  // Check if the most recent publish (if any) used a custom domain.
+  async _customDomain(shop) {
+    const custom = await ShopDomain.findOne({
+      where: { shopId: shop.id },
+      order: [['id', 'desc']]
+    })
+    return custom ? custom.domain.indexOf('ogn.app') > 0 : false
```
mikeshultz, line 141: I don't really think it has any benefit here, but you can use ILIKE like this with Sequelize:

    where: { domain: { [Op.iLike]: '%ogn.app' } }

No need to change.

backend/etl/processor.js
```diff
@@ -130,10 +132,62 @@ class EtlJobProcessor {
...
+      data.numDiscounts !== prevData.numDiscounts
+    ) {
+      return true
+    }
+
+    // Check if the shop was deployed.
+    const published = await ShopDeployment.findOne({
+      where: {
+        shopId: shop.id,
+        createdAt: { [Sequelize.Op.gte]: prevData.createdAt }
```
mikeshultz, line 170: Maybe you just want `Op.gt`?

mikeshultz (approved):
One comment, one question.  LGTM

backend/etl/processor.js
franckc: Ah yeah. Good to know. Thanks.

backend/etl/processor.js
franckc: Fixed.

Issue #502, opened 2020-08-26
Auto-select first payment option in checkout


nick:
When a buyer visits the payment screen on the checkout, the first payment option should be selected by default

theprivileges:
@nick could you clarify something for me, please?  Is the expected behavior to have the first payment on the list, regardless of the type of payment, to be `selected`, is that correct?

nick:
Yes that's correct :-)

PR #775, opened 2020-11-13
feat: Auto-select first payment option in checkout

theprivileges:
Add functionality to select the first payment option whenever the payment component mounts and the end-user has not made a selection on their own. Ensuring that once an end-user has picked an option that is the option that is selected should the end-user exit the _Payment_ page, and then return.

## Screenshots

Screen Recording 2020-11-12 at 21 07 28

Resolves: https://github.com/OriginProtocol/dshop/issues/502

### DIFF
--- a/shop/src/pages/checkout/ChoosePayment.js
+++ b/shop/src/pages/checkout/ChoosePayment.js
@@ -116,7 +116,7 @@ const ChoosePayment = () => {
   )
 
   useEffect(() => {
-    if (paymentMethods.length === 1 || paymentMethod == undefined) {
+    if (paymentMethods.length === 1 || paymentMethod === undefined) {
       dispatch({ type: 'updatePaymentMethod', method: paymentMethods[0] })
     }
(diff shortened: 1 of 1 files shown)


shop/src/pages/checkout/ChoosePayment.js
```diff
@@ -116,10 +116,10 @@ const ChoosePayment = () => {
   )
 
   useEffect(() => {
-    if (paymentMethods.length === 1) {
+    if (paymentMethods.length === 1 || paymentMethod === undefined) {
```
shahthepro, line 119: Can you also add a check to make sure paymentMethods isn't an empty array?

franckc (approved):
Looks good to me! Thanks @theprivileges for your work on this.

@shahthepro If that looks good to you as well, please merge in to master.

shahthepro:
Yeah, looks good to me. Thanks for the PR @theprivileges :)

PR #796, opened 2020-11-30
Wait out rate limit

shahthepro:
Printful seems to have [rate-limit](https://www.printful.com/docs) of 120 reqs/minute. This is causing failures during sync in some of the stores. This adds a sleep of 60s after every 75th request while getting data from printful API.

### DIFF
--- a/backend/scripts/printful/downloadProductData.js
+++ b/backend/scripts/printful/downloadProductData.js
@@ -51,8 +51,14 @@ async function downloadProductData({
     )
   }
 
+  let index = 1
   for (const id of ids) {
     await getProduct({ PrintfulURL, apiAuth, OutputDir, id })
+    if (index % 75 === 0) {
+      // Wait out rate limit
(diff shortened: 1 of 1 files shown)


shahthepro:
Okay, this throws Gateway Timeout error after a minute on frontend, however the sync will still be in progress in background. It didn't throw on rate-limiting errors on backend. Will work on moving this sync to the existing queue too to avoid the Gateway Timeout errors.

franckc (commented):
LGTM - thanks for adding this.

Couple minor comments:
 - we should probably move this method out of scripts since it is being called by the sync queue. Maybe under backend/src/logic/printful?
 - mention the rate limiting in the jsdoc of the downloadProductData so that the callers in the future don't get surprised if the method takes a while
 - consider adding a log.info statement "sleeping for 60sec"

shahthepro:
Sure, will look in it

shahthepro:
@franckc Have addressed your comments in this PR: #800

Issue #780, opened 2020-11-13
Token decimal support


nick:
The USDC token does not use 18 decimals, which the checkout currently assumes when checking balances. This results in an 'insufficient balance' error shown to the user.

franckc:
As reported by Max, this also impact the TrustToken (TRU) since it has 8 decimals. 
See https://etherscan.io/address/0x4c19596f5aaff459fa38b0f7ed92f11ae6543784#readProxyContract

kayyoo23:
Thanks team. 👍  Let me know when our target ETA is and I will let TrustToken team know. They are ready to launch their Dshop whenever this issue is fixed.

PR #797, opened 2020-11-30
Add token decimal support

shahthepro:
Related issue: #780

### DIFF
--- a/shop/src/components/payment/crypto/Crypto.js
+++ b/shop/src/components/payment/crypto/Crypto.js
@@ -24,6 +24,8 @@ const waitForAllowance = ({ wallet, marketplace, amount, token }) => {
             address,
             marketplace.address
           )
+          // TODO: formatUnits returns string, Evaluate doing
+          // parseFloat before the comparison
           const allowanceEth = ethers.utils.formatUnits(allowance, decimals)
           if (allowanceEth >= amount) {
             resolve()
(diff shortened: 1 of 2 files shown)


shop/src/components/payment/crypto/Crypto.js
```diff
@@ -17,18 +17,22 @@ import TokenList from './_TokenList'
...
-          })
-          .catch(reject)
+      async function checkAllowance() {
+        try {
+          const decimals = await token.contract.decimals()
+          const allowance = await token.contract.allowance(
+            address,
+            marketplace.address
+          )
+          const allowanceEth = ethers.utils.formatUnits(allowance, decimals)
```
franckc, line 27: Doesn't ethers.utils.formatUnits return a string?
Should it get converted to a number before doing the comparison below?

franckc (approved):
LGTM aside from 1 comment below. PTAL before merging.

shop/src/components/payment/crypto/Crypto.js
shahthepro: Yeah, it does return a finite number wrapped as string. But the comparisons works because it's JS 😅 

I'm not sure if parsing it as number will break anything else, will make a note of it and change it in a different branch.

shahthepro:
Merging this in now, will work on that open comment in a different branch

PR #803, opened 2020-12-02
GH Action tests for backend

mikeshultz:
Adds GH action to run backend tests and fixes some things along the way.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -1,4 +1,4 @@
-name: Backend Tests
+name: Backend
 
 on:
   push:
@@ -40,6 +40,10 @@ jobs:
 
       - name: Install dependencies
         run: yarn
+
+      - name: Build frontend distribution for backend
+        run: yarn run build:dist
+        working-directory: ./shop
 
       - name: Set LOG_LEVEL
         run: echo "LOG_LEVEL=ERROR" >> $GITHUB_ENV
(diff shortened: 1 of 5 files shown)


.github/workflows/backend.yml
```diff
@@ -0,0 +1,53 @@
...
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies
+        run: yarn
+
+      - name: Build frontend distribution for backend
+        run: yarn run build:dist
```
franckc, line 45: This is a slow step. Is that necessary?
It takes a while to build (several minutes on my host).

franckc (approved):
Great! Thanks for adding this.

One question inline otherwise LGTM

.github/workflows/backend.yml
mikeshultz: Yeah... Some of the tests reliy on having the shop build available.  We might be able to optimize later by adding this as artifacts.  I don't think the tests really care that the build is up to date.

franckc:
Maybe for the tests we only build 1 theme by passing it as an argument to the script?
https://github.com/OriginProtocol/dshop/blob/master/shop/buildThemes.sh#L31

Issue #15, opened 2020-05-11
UI should indicate user can't pay with ERC20 if ETH balance is 0


gullible1:
The user had sufficient OGN to cover the cost of the purchase but didn't have any ETH, and was unable to confirm the puchase.

The UI should indicate that the user can't pay with ERC20 if ETH balance in wallet is 0.

![Screen_Shot_2020-05-08_at_3 07 07_PM](https://user-images.githubusercontent.com/46481405/81517414-79a81b80-936d-11ea-90e7-d2f9fee14297.png)

franckc:
Not a bug. You need a bit of ETH to pay for gas fees. Min tested again and confirmed it works once she had ETH in her wallet.

tomlinton:
UI should explain that they have no ETH and so can't pay even though they have sufficient OGN. Will change tag.

gullible1:
Thanks Tom, appreciate it (:

phyninja:
Hi @gullible1,

With help from @shahthepro and @franckc, I have been able to observe the user flow leading up to this 'issue' first hand. Hopefully I will have more to share in the days ahead. 

Best,
Rajath

phyninja:
### Clarification request

I'm informed that the point here is for the DShop DApp to confirm that there exists a positive ETH balance in the user's wallet. At what point should the DApp run such a query? 

The screenshots below document the steps leading up to payment with any cryptocurrency, from my perspective :
- #### **Step 1: User arrives at the payment page. They enable their wallet by clicking on the Metamask extension (top right)**
![Step 1](https://user-images.githubusercontent.com/10854442/99536334-3986bd00-29d0-11eb-88e4-ab5c7f0a4040.png)

<br>

- #### **Step 2: User selects a token on the DShop UI and clicks 'Pay...'**
![Step 2](https://user-images.githubusercontent.com/10854442/99538195-997e6300-29d2-11eb-93c0-53a08542c8e1.png)

<br>

- #### **Step 3: Control is transferred to Metamask. Metamask UI updates, asking the user for a final confirmation**
![Step 3](https://user-images.githubusercontent.com/10854442/99536350-3e4b7100-29d0-11eb-9a5f-f745eed033d5.png)

phyninja:
[...]

PR #816, opened 2020-12-05
Response to #15 - UI should indicate user can't pay with ERC20 if ETH balance is 0

phyninja:
UI now displays the following message when a user doesn't have any ETH in their connected account:
"You will need some ETH to pay for gas fees. Please deposit some ether and refresh the page to continue." 

Screenshots: https://github.com/OriginProtocol/dshop/issues/15#issuecomment-733640860


shop/src/components/payment/crypto/CryptoDirect.js
```diff
@@ -18,6 +18,9 @@ const PayWithCryptoDirect = ({ submit, encryptedData, onChange, loading }) => {
   const [{ cart }, dispatch] = useStateValue()
   const [activeToken, setActiveToken] = useState({})
   const [tokenPrice, setTokenPrice] = useState('')
+  const [walletBalance, setWalletBalance] = useState({
+    isZero: false
+  }) //create an object called walletBalance, assuming for now that the buyer has some ETH in it already. Credit: https://www.geeksforgeeks.org/reactjs-usestate-hook/
```
shahthepro, line 23: While I appreciate you adding comments, I'd suggest to avoid adding links to generic tutorials. Links to API reference is fine. But this seems out of place.

shop/src/components/payment/crypto/CryptoDirect.js
```diff
@@ -92,22 +95,44 @@ const PayWithCryptoDirect = ({ submit, encryptedData, onChange, loading }) => {
...
+          setWalletBalance({
+            isZero: res.isZero() //boolean to check whether balance is zero. Reference: https://docs.ethers.io/v5/api/utils/bignumber/
+          })
+        )
+        .catch((err) => console.log('Failed to query wallet balance.\n' + err))
+    )
+  }, [wallet.address, activeToken.id, cryptoSelected])
+  // The 'useEffect' hook above verifies that the buyer's wallet has a positive ETH balance,
+  // and runs every time they change their connected crypto account, the payment token of choice, or the 'Cryptocurrency' payment option.
+  // Credit: https://dmitripavlutin.com/react-useeffect-explanation/
```
shahthepro, line 115: Same comment about this link as above.

shop/src/components/payment/crypto/CryptoDirect.js
```diff
@@ -92,22 +95,44 @@ const PayWithCryptoDirect = ({ submit, encryptedData, onChange, loading }) => {
 
   const acceptedTokens = config.acceptedTokens
 
+  useEffect(() => {
+    async function queryWalletBalance() {
+      return await wallet.signer.getBalance()
+    } //returns a promise that, if resolved, will give the balance of the wallet in wei (BigNumber)
```
shahthepro, line 101: If you only need this function to return promise, you remove the `async` and `await` to keep code simple. It'll still return a promise that you can `await`

shop/src/components/payment/crypto/CryptoDirect.js
```diff
@@ -92,22 +95,44 @@ const PayWithCryptoDirect = ({ submit, encryptedData, onChange, loading }) => {
...
+
+    onChange(
+      queryWalletBalance()
```
shahthepro, line 104: Why is it in a separate function, can't we just use `wallet.signer.getBalance().then(...)`?

shahthepro (approved):
LGTM. Thanks for making these changes @phyninja

PR #842, opened 2020-12-15
Fixes custom domain handling with CDN deployments

### LINKED ISSUE #835: SSL cert missing domains with Cloud CDN on deploy
Seems that deployments occasionally create a new cert with only the default domain, causing cert issues on any custom domains.   Not sure of the conditions yet.

This can be worked around by deleting and reconfiguring the domain after the deployment.

This currently only impacts the Origin store

mikeshultz:
Brings domain handling in line between the custom domain API endpoints and the deployment workflow.

Fixes #835


franckc (approved):
LGTM

Issue #833, opened 2020-12-13
Review the use of variants.id as a unique marker for a variant 


phyninja:
A brief look at the codebase showed a few places where the use of **Variant ID** needs inspection because that field currently does not uniquely identify a variant. Example:

![Screenshot 2020-12-13 at 12 36 25 PM](https://user-images.githubusercontent.com/10854442/102005718-1508d100-3d41-11eb-9d6f-b3180c2f0f03.png)

phyninja:
Related code from _dshop/shop/src/utils/generateVariants.js_:
<pre>
export const generateVariants = (product) => {
  const { options, availableOptions, variants } = product

  if (!options || !availableOptions) return []
  if (!options.length || !availableOptions.length) return []

  // Map existing variants for easy access and carrying over values
  const existingVairants = (variants || []).reduce((obj, variant) => {
    return {
      ...obj,
      [variant.options.join('|||')]: variant
    }
  }, {})

  const newVariants = getAllCombinations(...availableOptions).map(
    (optionCombo, index) => {
      const comboTitle = `${product.title} - ${optionCombo.join(' / ')}`

      return {
        id: index,
        title: comboTitle,
        name: comboTitle,
        price: product.price,
        image: product.image,
        available: true,

        <b>// Default to previous values if it exists</b>
        ...existingVairants[optionCombo.join('|||')],

        // Set {option1, option2, ...} values
        ...optionCombo.reduce(
          (optsObj, opt, optIndex) => ({
            ...optsObj,
            [`option${optIndex + 1}`]: opt
          }),
          {}
        ),

        // Set options array
        options: optionCombo
      }
    }
  )

  return newVariants
}
</pre>

PR #848, opened 2020-12-16
Resetting Variant ID when new variants are generated

phyninja:
I have not been able to replicate the behavior mentioned in #833 after making the change attached.

### DIFF
--- a/shop/src/utils/generateVariants.js
+++ b/shop/src/utils/generateVariants.js
@@ -29,7 +29,7 @@ export const generateVariants = (product) => {
   if (!options.length || !availableOptions.length) return []
 
   // Map existing variants for easy access and carrying over values
-  const existingVairants = (variants || []).reduce((obj, variant) => {
+  const existingVariants = (variants || []).reduce((obj, variant) => {
     return {
       ...obj,
       [variant.options.join('|||')]: variant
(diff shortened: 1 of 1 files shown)


shahthepro (commented):
Thanks for finding and fixing this bug. Is this PR ready to be merged? Or do you want to take some time doing more tests?

phyninja:
If afforded the luxury, it would be best to verify that this doesn't break anything on the back end.

phyninja:
@shahthepro
Update: 
My ignorance about the Printful platform disallows an understanding of the code in dshop/backend/logic/printful/sync/writeProductData.js

I have reviewed the use of variant.id in the following files, and am okay with this change in relation to them:
dshop/shop/src/pages/Product.js
dshop/backend/logic/products/index.js
dshop/shop/src/themes/ybm/components/Product.js
dshop/shop/src/utils/useProduct.js
dshop/shop/src/utils/useAdminProduct.js
dshop/shop/src/utils/useProducts.js

phyninja:
I'm ready for this PR to be merged. My belief is that generateVariants.js only applies to products managed via the DShop platform. Furthermore, [DShop's welcome packet](https://drive.google.com/file/d/1_O8nylYyNh23O9vSnWmkIniiMOXSz9Ql/view) mentions:
> Dshop does not support both Printful and non-Printful products at the same time. If there are non-Printful products in the store, sync-ing a Printful account will cause those products to be lost.

shahthepro:
Thanks @phyninja, Yeah, we don't override variant ID of printful products. It must be unique already.

Issue #320, opened 2020-07-28
Show icons for custom tokens


joshfraser:
Low priority, but it would be nice to show the proper icons when people add a custom token as an accepted payment method. 

[Metamask](https://github.com/MetaMask/eth-contract-metadata/blob/master/contract-map.json), [TrustWallet](https://github.com/trustwallet/tokens/tree/master/tokens), and [Ethereum Lists](https://github.com/ethereum-lists/tokens/tree/master/tokens/eth) all maintain popular lists that we could use.

phyninja:
Hi @joshfraser,

A draft pull request has been created to give one a sense of what the checkout page might look like if we used the assets managed by TrustWallet: https://github.com/OriginProtocol/dshop/pull/852
Feedback is most welcome.

PR #852, opened 2020-12-22
Token image support

phyninja:
Feature requested in #320 

The screenshot below is an approximate rendering of the change:
Screenshot 2020-12-22 at 6 14 22 PM

### DIFF
--- a/shop/src/components/payment/crypto/_TokenList.js
+++ b/shop/src/components/payment/crypto/_TokenList.js
@@ -41,7 +41,7 @@ const TokenList = ({
                   onChange={() => setActiveToken(token)}
                 />
                 <div className={`token-logo${isActive ? ' active' : ''}`}>
-                  <img src={`images/payment/${token.name.toLowerCase()}.svg`} />
+                  <img src={`https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/${token.address}/logo.png`} />
                 </div>
                 <div>{token.name}</div>
               </td>
--- a/shop/src/components/payment/crypto/_TokenList.js
+++ b/shop/src/components/payment/crypto/_TokenList.js
@@ -3,6 +3,8 @@ import fbt from 'fbt'
 
 import usePrice from 'utils/usePrice'
 
+const Web3Utils = require('web3-utils')
+
 const TokenList = ({
   acceptedTokens,
   activeToken,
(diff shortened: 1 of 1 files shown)


nick:
Hrm I don't think we should rely on github to serve this content as it's introducing another dependency. Might need to rethink how we approach this

phyninja:
Hi @nick,

I had the time to look at this from the perspective of [Ethereum Lists ](https://github.com/ethereum-lists/tokens/tree/master/tokens/eth); here's what I learned:
1) There are currently about 1958 tokens listed there, but about two-thirds of them don't have any information about images. The (751) tokens that have their "logo" field populated have been compiled into this list: 
[EthereumListsTokenCollection.xlsx](https://github.com/OriginProtocol/dshop/files/5741895/EthereumListsTokenCollection.xlsx)
2) Between the tokens in the list above and the 6064 token images hosted using [TrustWallet's github repository](https://github.com/trustwallet/assets/tree/master/blockchains/ethereum), I found some in common; such tokens can be identified by the field **Logo Alt Source**.

phyninja:
Reading about this a little more, I see that Metamask favors EIP-747's idea that users should be allowed to add token information (to their wallet) themselves; this may include URLs to the token image, and it seems like the argument here is that storing this information on the client side (with the provider being blind to it) is a step in the direction of (more) decentralization. Other considerations include
- The fact that managing a list of assets is a "security-delicate" task with significant "administrative overhead", and that
- List maintainers may face political pressure by asset creators to list the latter's "otherwise unknown" assets on their platform

**Sources**:
https://github.com/MetaMask/contract-metadata#metamaskcontract-metadata
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-747.md#rationale

phyninja:
@nick if we implemented an approach in which DShop would try to serve the image from say, Github, failing which it would try a 'fallback' URL, could we mitigate the dependency issue?

phyninja:
I'm checking in on this since it's been a few weeks since there has been an update. 
@nick, when possible, could you please respond to the question above?

phyninja:
@nick,
Your concern about the centralization of token images is a shared one (see [EIP 2569](https://eips.ethereum.org/EIPS/eip-2569)). Given the uncertainty of the timeline in which we may see that proposal come to fruition, however, I'm advocating for us to implement the approach I previously mentioned. 

@shahthepro 
Would you please review the commits that follow this comment, and recommend changes if any?

Best,
Rajath

PR #859, opened 2020-12-28
Revert "Response to #15 - UI should indicate user can't pay with ERC20 if ETH balance is 0"

shahthepro:
Reverts OriginProtocol/dshop#816


shahthepro:
@phyninja I have reverted your change since it breaks on prod when `wallet.signer` is undefined

Issue #814, opened 2020-12-04
Printful stock management


nick:
We need to mark items as 'out of' or 'in' stock when the stock_updated [webhook](https://www.printful.com/docs/webhooks) is called from Printful

PR #870, opened 2021-01-15
Zahupro/printful stock update

iam-hussain:
Related: https://github.com/OriginProtocol/dshop/issues/814

-  Tested through postman and https://www.printful.com/docs/webhook-simulator

### DIFF
--- a/backend/logic/printful/sync/stockUpdate.js
+++ b/backend/logic/printful/sync/stockUpdate.js
@@ -1,5 +1,4 @@
 const _get = require('lodash/get')
-const program = require('commander')
 const { decryptConfig } = require('../../../utils/encryptedConfig')
 
 const { get } = require('../api')
@@ -7,11 +6,10 @@ const { Product } = require('../../../models')
 
 /**
  * Updates discontinued variants and variants that are out of stock to a collection.
- * And creates collection if it doesn't exist
  *
- * @param {Object} shop
- * @param {Number|String} productId
- * @param {Array<Object>} variants
+ * @param {Object} shop: Shop DB object
+ * @param {Number|String} productId: External ID of the product
+ * @param {Array<Object>} variants: List of variants that are out of stock/discontinued
  */
 const stockUpdate = async (shop, productId, variants) => {
   const shopConfig = decryptConfig(shop.config)
--- a/backend/logic/printful/sync/stockUpdate.js
+++ b/backend/logic/printful/sync/stockUpdate.js
@@ -1,5 +1,4 @@
 const _get = require('lodash/get')
-const program = require('commander')
 const { decryptConfig } = require('../../../utils/encryptedConfig')
 
 const { get } = require('../api')
@@ -32,7 +31,8 @@ const stockUpdate = async (shop, productId, variants) => {
       const productData = await get(`/store/products/${productId}`, {
         apiKey: apiKey
       })
-      existingVariants = _get(productData, 'result.sync_product.synced', 0) - variants.length
+      existingVariants =
+        _get(productData, 'result.sync_product.synced', 0) - variants.length
       existingVariants = existingVariants < 0 ? 0 : existingVariants
     }
 
--- a/backend/logic/printful/sync/stockUpdate.js
+++ b/backend/logic/printful/sync/stockUpdate.js
@@ -1,0 +1,66 @@
+const get = require('lodash/get')
+const {
+    decryptConfig
+} = require('../../../utils/encryptedConfig')
+
+const {
+    Product
+} = require('../../../models')
+const {
+    getLogger
+} = require('../../../utils/logger')
+
+const log = getLogger('utils.printful.writeProductData')
+
+/**
+ * Appends a productId to a collection.
+ * And creates collection if it doesn't exist
+ *
+ * @param {Object} shop
+ * @param {Number|String} productId
+ * @param {Array<Object>} variants
+ */
+const stockUpdate = async (shop, productId, variants) => {
+    try {
+        const shopConfig = decryptConfig(shop.config)
+        const shopInventory = get(shopConfig, 'inventory')
+        if (shopInventory) {
+            // Add to products model
+            const variantsStock = variants.reduce(
+                (stockData, variant) => ({
+                    ...stockData,
+                    [variant]: 0
+                }), {}
+            )
+
+            const updatedStockData = {
+                productId: productId.toString(),
+                shopId: shop.id,
+                stockLeft: 0,
+                variantsStock
+            }
+
+            let existingDbItem
+
+            if (productId) {
+                existingDbItem = await Product.findOne({
+                    where: {
+                        productId: productId.toString(),
+                        shopId: shop.id
+                    }
+                })
+            }
+
+            if (existingDbItem) {
+                await existingDbItem.update(updatedStockData)
+            } else {
+                await Product.create(updatedStockData)
+            }
+        }
+    } catch (e) {
+        log.error('Failed to stock update while syncing printful products', e)
+    }
+    return
+}
+
+module.exports = stockUpdate
--- a/backend/logic/printful/sync/stockUpdate.js
+++ b/backend/logic/printful/sync/stockUpdate.js
@@ -1,13 +1,12 @@
-const get = require('lodash/get')
+const _get = require('lodash/get')
+const program = require('commander')
 const { decryptConfig } = require('../../../utils/encryptedConfig')
 
+const { get } = require('../api')
 const { Product } = require('../../../models')
-const { getLogger } = require('../../../utils/logger')
-
-const log = getLogger('utils.printful.writeProductData')
 
 /**
- * Appends a productId to a collection.
+ * Updates discontinued variants and variants that are out of stock to a collection.
  * And creates collection if it doesn't exist
  *
  * @param {Object} shop
@@ -15,45 +14,51 @@ const log = getLogger('utils.printful.writeProductData')
  * @param {Array<Object>} variants
  */
 const stockUpdate = async (shop, productId, variants) => {
-  try {
-    const shopConfig = decryptConfig(shop.config)
-    const shopInventory = get(shopConfig, 'inventory')
-    if (shopInventory) {
-      // Add to products model
-      const variantsStock = variants.reduce(
-        (stockData, variant) => ({
-          ...stockData,
-          [variant]: 0
-        }),
-        {}
-      )
+  const shopConfig = decryptConfig(shop.config)
+  const shopInventory = _get(shopConfig, 'inventory')
+  if (shopInventory) {
+    const variantsStock = variants.reduce(
+      (stockData, variant) => ({
+        ...stockData,
+        [variant]: 0
+      }),
+      {}
+    )
 
-      const updatedStockData = {
-        productId: productId.toString(),
-        shopId: shop.id,
-        stockLeft: 0,
-        variantsStock
-      }
+    const apiKey = shopConfig.printful
+    let existingVariants = 0
 
-      let existingDbItem
-
-      if (productId) {
-        existingDbItem = await Product.findOne({
-          where: {
-            productId: productId.toString(),
-            shopId: shop.id
-          }
-        })
-      }
-
-      if (existingDbItem) {
-        await existingDbItem.update(updatedStockData)
-      } else {
-        await Product.create(updatedStockData)
-      }
+    if (apiKey) {
+      const productData = await get(`/store/products/${productId}`, {
+        apiKey: apiKey
+      })
+      existingVariants = _get(productData, 'result.sync_product.synced', 0) - variants.length
+      existingVariants = existingVariants < 0 ? 0 : existingVariants
     }
-  } catch (e) {
-    log.error('Failed to stock update while syncing printful products', e)
+
+    const updatedStockData = {
+      productId: productId.toString(),
+      shopId: shop.id,
+      stockLeft: existingVariants * 10000,
+      variantsStock
+    }
+
+    let existingDbItem
+
+    if (productId) {
+      existingDbItem = await Product.findOne({
+        where: {
+          productId: productId.toString(),
+          shopId: shop.id
+        }
+      })
+    }
+
+    if (existingDbItem) {
+      await existingDbItem.update(updatedStockData)
+    } else {
+      await Product.create(updatedStockData)
+    }
   }
   return
 }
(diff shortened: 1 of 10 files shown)


backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,61 @@
+const get = require('lodash/get')
+const { decryptConfig } = require('../../../utils/encryptedConfig')
+
+const { Product } = require('../../../models')
+const { getLogger } = require('../../../utils/logger')
+
+const log = getLogger('utils.printful.writeProductData')
+
+/**
+ * Appends a productId to a collection.
```
franckc, line 10: This description seems incorrect, right? This method should only update stockLeft in the Product table?

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,61 @@
...
+            productId: productId.toString(),
+            shopId: shop.id
+          }
+        })
+      }
+
+      if (existingDbItem) {
+        await existingDbItem.update(updatedStockData)
+      } else {
+        await Product.create(updatedStockData)
```
franckc, line 52: Under which scenario would we want to create a new product?

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,61 @@
...
+          ...stockData,
+          [variant]: 0
+        }),
+        {}
+      )
+
+      const updatedStockData = {
+        productId: productId.toString(),
+        shopId: shop.id,
+        stockLeft: 0,
```
franckc, line 34: My understanding is that stockLeft is the cumulative stock left across all the variants.
When this method gets called is stockLeft always completely depleted? Or does it gets calls every time a variant's stock changes?

backend/logic/printful/webhook.js
```diff
@@ -207,18 +224,17 @@ const processShippedEvent = async (event, shopId) => {
  * Processes a ProductUpdated event sent by Printful via webhook.
  * @param event
  * @param shopId
+ * @param shopData
```
franckc, line 227: nit: document the type for shopData (and the other params while we are at it).

backend/logic/printful/webhook.js
```diff
@@ -237,10 +253,42 @@ const processUpdatedEvent = async (event, shopId) => {
...
+      const stockDiscontinuedVariant = get(
+        event,
+        'variant_stock.discontinued',
+        []
+      )
+
+      await stockUpdate(
+        shop,
+        product_id,
+        uniq([...stockOutVariant, ...stockDiscontinuedVariant])
```
franckc, line 278: Should we test for this list to not be empty before calling stockUpdate?
Or handle that in the stockUpdate method

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,61 @@
...
+        })
+      }
+
+      if (existingDbItem) {
+        await existingDbItem.update(updatedStockData)
+      } else {
+        await Product.create(updatedStockData)
+      }
+    }
+  } catch (e) {
```
franckc, line 55: Not sure we need a try/catch in this method since the caller already calls it within a try/catch.

Also, we very rarely check the system logs. So we'll probably not notice any error. It's good to leave that logging but in addition, it would probably make sense to log the error in Sentry since we have it configured to end us notifications by email in case of failure.

backend/queues/printfulSyncProcessor.js
```diff
@@ -69,6 +71,12 @@ const processor = async (job) => {
 
   let taskStartTime = Date.now()
 
+  const shop = await Shop.findOne({ where: { id: shopId } })
+  const shopConfig = decryptConfig(shop.config)
+  const network = await Network.findOne({ where: { active: true } })
+  const netConfig = getConfig(network.config)
+  const backendUrl = process.env.BACKEND_TUNNEL || netConfig.backendUrl
```
franckc, line 78: What is BACKEND_TUNNEL? Is that a new env var?
If we want to keep it, let's document its usage in the README.

franckc (commented):
Thanks for working on this! A few questions inline, PTAL...

backend/logic/printful/sync/stockUpdate.js
ZaHuPro: > When this method gets called is stockLeft always completely depleted? Or does it gets calls every time a variant's stock changes?

It only sends the out of stock and discontinued variants. It does not send the remaining stock for other variants. It could be a problem if only some variants are out of stock. I can compare variants array to find in stock items and set the stock to a higher number like 10000.

backend/logic/printful/sync/stockUpdate.js
ZaHuPro: It gets created only if the data does not exist in the DB. Otherwise, it updates it

backend/logic/printful/webhook.js
ZaHuPro: It should be called even if the array is empty, but the `stockUpdate` function should update  `stockLeft` accordingly.

from printful Docs
> Contains product id and ids of it's discontinued variants and variants that are out of stock. Variant ids that are not present should be considered as active and in stock.

https://www.printful.com/docs/webhooks

backend/logic/printful/sync/stockUpdate.js
ZaHuPro: Okay, I will remove it

backend/queues/printfulSyncProcessor.js
ZaHuPro: It's just for testing locally, I will add it to README.

backend/logic/printful/sync/stockUpdate.js
ZaHuPro: My bad, I will update

backend/logic/printful/webhook.js
ZaHuPro: Okay, I will update

backend/logic/printful/sync/stockUpdate.js
ZaHuPro: Updated

backend/logic/printful/sync/stockUpdate.js
ZaHuPro: Updated

backend/logic/printful/webhook.js
ZaHuPro: Updated

backend/logic/printful/sync/stockUpdate.js
ZaHuPro: Removed

backend/queues/printfulSyncProcessor.js
ZaHuPro: Added in README

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,66 @@
...
+          productId: productId.toString(),
+          shopId: shop.id
+        }
+      })
+    }
+
+    if (existingDbItem) {
+      await existingDbItem.update(updatedStockData)
+    } else {
+      await Product.create(updatedStockData)
```
franckc, line 60: Sorry, I'm not very familiar with that area of our codebase, but I still don't understand exactly why we need to create a Product row here. Under which scenario would we need this? I would expect that for Printful to call us back on this webhook, a product row must have already been created already, earlier on when the product was added to the store in the first place?
If that holds, then I'd consider receiving a webhook call for a product row that does not exit to be an error that we log but we don't create a product database row.

Also note that adding Product rows in the database is not enough to add products to the store. For that to happen, static json files also need to be stored on disk under the store's data directory. See code from the upserProduct method [here](https://github.com/OriginProtocol/dshop/blob/master/backend/logic/products/index.js#L353).

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,66 @@
...
+const { get } = require('../api')
+const { Product } = require('../../../models')
+
+/**
+ * Updates discontinued variants and variants that are out of stock to a collection.
+ * And creates collection if it doesn't exist
+ *
+ * @param {Object} shop
+ * @param {Number|String} productId
+ * @param {Array<Object>} variants
```
franckc, line 14: It would be helpful to document that variants is the list of variants that are out of stock or discontinued.

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,66 @@
...
+        [variant]: 0
+      }),
+      {}
+    )
+
+    const apiKey = shopConfig.printful
+    let existingVariants = 0
+
+    if (apiKey) {
+      const productData = await get(`/store/products/${productId}`, {
```
franckc, line 32: Why do we need to fetch the product data from Printful? Please add in the code to comment to explain. This is not obvious.

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,66 @@
...
+        apiKey: apiKey
+      })
+      existingVariants = _get(productData, 'result.sync_product.synced', 0) - variants.length
+      existingVariants = existingVariants < 0 ? 0 : existingVariants
+    }
+
+    const updatedStockData = {
+      productId: productId.toString(),
+      shopId: shop.id,
+      stockLeft: existingVariants * 10000,
```
franckc, line 42: Add a comment as to why we do his math.

Overall I'm still unsure this is the best way to go.

Not being super familiar with the overall Printfu sync flow, my naive thoughts would be that at a high level the logic would be as follow:
  1. Printful calls this webhook with a list of variants that are out of stock or discontinued.
  2. We load the corresponding Product row from the database
  3. We update the field Product.variantsStock to set quantity to zero for all variants  out of stock/discontinued
  4. We update the field Product.stockLeft by recomputing the sum of all variants in Product.variantsStock

franckc (commented):
My apologies for the super long turnaround on this review :(

Some comments inline. Please take a look.

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,66 @@
+const _get = require('lodash/get')
+const program = require('commander')
+const { decryptConfig } = require('../../../utils/encryptedConfig')
+
+const { get } = require('../api')
```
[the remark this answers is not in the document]
ZaHuPro: I didn't rename it, I just imported it as is

backend/logic/printful/sync/stockUpdate.js
ZaHuPro: > I would expect that for Printful to call us back on this webhook, a product row must have already been created already, earlier on when the product was added to the store in the first place?

The DB items are not created in the sync
https://github.com/OriginProtocol/dshop/blob/cc6782b0c0e989feee9bc35bbf5d2c7ee37a7db4/backend/logic/printful/sync/writeProductData.js#L52

> Also note that adding Product rows in the database is not enough to add products to the store. 

I am not adding the new product here, the table contains only stock data

backend/logic/printful/sync/stockUpdate.js
ZaHuPro: I need to get all the variant's stock data because printful only sends out of stock and discontinued items. I will add that in the comments

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,66 @@
...
+const { Product } = require('../../../models')
+
+/**
+ * Updates discontinued variants and variants that are out of stock to a collection.
+ * And creates collection if it doesn't exist
```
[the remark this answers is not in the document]
ZaHuPro: Removed

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,66 @@
+const _get = require('lodash/get')
+const program = require('commander')
```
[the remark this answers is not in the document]
ZaHuPro: removed

backend/logic/printful/sync/stockUpdate.js
ZaHuPro: Updated

backend/logic/printful/sync/stockUpdate.js
franckc: ok got it now. I did not realize Printful Sync does not pull the inventory data.sounds good.

backend/logic/printful/sync/stockUpdate.js
franckc: I now understand Printful Sync does not send inventory data. makes sense then.

backend/logic/printful/sync/stockUpdate.js
franckc: Cleared this up on discord..

backend/logic/printful/webhook.js
franckc: got it. thanks.

backend/logic/printful/sync/stockUpdate.js
```diff
@@ -0,0 +1,65 @@
...
+      })
+      existingVariants =
+        _get(productData, 'result.sync_product.synced', 0) - variants.length
+      existingVariants = existingVariants < 0 ? 0 : existingVariants
+    }
+
+    const updatedStockData = {
+      productId: productId.toString(),
+      shopId: shop.id,
+      stockLeft: existingVariants * 10000,
```
franckc, line 41: This will work but seems a bit hacky:)

In case existingVariants is not zero, as opposed to using a very large value, how about using -1 to indicate the inventory is unlimited (well, until Printful notifies us it is out of stock)?

It would require to update the places in the code where we decrement stockLeft to handle -1 as a special value. But it seems more readable and would also allow to easily query those Products that have unlimited inventory in the DB?

backend/queues/printfulSyncProcessor.js
```diff
@@ -83,6 +91,16 @@ const processor = async (job) => {
   )
   taskStartTime = Date.now()
 
+  if (updatedIds.length > 0) {
```
franckc, line 94: nit: it makes sense. but for readability, add a comment explaining why we need to re-register the printful hook.

franckc (approved):
Please take a look at my comment below about setting stockLeft to -1 rather than a huge value.
A few small nits inline.
Also it would be great to add some unit test coverage. See backend/test/inventory.test.js

Otherwise LGTM.

backend/logic/printful/sync/stockUpdate.js
franckc: @nick Jakir is trying to wrap up this PR. You are more familiar than I am on the Printful inventory management... How do you feel about my suggestion to use -1 for stockLeft? Or is there a better way in your opinion to handle this?

PR #888, opened 2021-02-02
Printful availability

nick:


backend/logic/order/index.js
```diff
@@ -541,7 +541,9 @@ async function updateInventoryData(
 
   const quantModifier = increment ? 1 : -1
 
-  const cartItems = get(cartData, 'items', [])
+  const cartItems = get(cartData, 'items', []).filter(
```
franckc, line 544: nit: add a comment explaining we filter out items for which the inventory is managed externally, such as Printful items.

backend/scripts/printful/getSizeGuide.js
```diff
@@ -1,5 +1,8 @@
-const fetch = require('node-fetch')
 const fs = require('fs')
```
franckc, line 1: nit: add a comment explaining what this script is for and how to use it. same thing for script printful2.js below.

franckc (approved):
Nice! Minor comments inline otherwise LGTM. Go NFT project :)

PR #890, opened 2021-02-03
Fix #883

### LINKED ISSUE #883: Printful auto-fulfillment failures due to invalid variant id
I investigated an order auto-fulfillment issue reported by a merchant with a printful enabled shop (shop id 1531).

A few of their orders (for example: J25XTMME and HLBR9QG8) failed to auto-fullfill.
The reason seems to be that in the data dir on disk, for some (but not all) products, the data.json file has variant ids that are not present in the printful-ids.json file.

For example, the order J25XTMME includes the product `whale-line-tee`.
Here is an excerpt of `whale-line-tee/data.json`:
```
  "variants": [
    {
      "title": "Whale Line Tee - Black / S",
      "name": "Whale Line Tee - Black / S",
      "price": 2700,
      "image": "958f65f529bc81af5535074fc9f80ea8.jpg",
      "available": true,
      "id": 0,
      "externalId": 2396252583,
      "option1": "Black",
      "option2": "S",
      "option3": null,
      "options": [
        "Black",
        "S"
      ]
    },
```
Note how **the variant has id 0**.

But according to the printful-ids.json file, the variant id should actually be a 4 digits number. 2805 in this specific case:
```
  "whale-line-tee": {
    "2801": 2396252620,
    "2802": 2396252621,
    "2803": 2396252622,
    "2804": 2396252623,
    "2805": 2396252583,
    "2806": 2396252584,
    "2807": 2396252585,
    "2808": 2396252586,
    "2809": 2396252593,
    "2810": 2396252594,
    "2811": 2396252595,
    "2812": 2396252596,
    "2817": 2396252588,
    "2818": 2396252589,
    "2819": 2396252590,
    "2820": 2396252591,
    "2901": 2396252598,
    "2902": 2396252599,
    "2903": 2396252600,
    "2904": 2396252601,
    "2977": 2396252603,
    "2978": 2396252604,
    "2979": 2396252605,
    "2980": 2396252606,
    "3060": 2396252624,
    "3061": 2396252587,
    "3062": 2396252597,
    "3064": 2396252592,
    "3084": 2396252602,
    "3103": 2396252610
  },
```

As a result, the id lookup when generating the Prinful order fails. See this [line](https://github.com/OriginProtocol/dshop/blob/master/packages/utils/generatePrintfulOrder.js#L28).

I'm not sure what could cause this inconsistency between printful-ids.json and the product's data.json?

@nick or @shahthepro 
 - Any idea about the root cause?
 - Would triggering a manual Printful Sync repare those files as a temporary workaround until we figure the root cause?

As a side note, I'm also unclear as to why the code does that id lookup in the generatePrintfulOrder method rather than directly using the externalVariantId which is available in the order...

shahthepro:
- Skip overriding variant IDs for externally managed products (printful)
- Add a script to fix already broken variant IDs

### DIFF
--- a/shop/src/utils/generateVariants.js
+++ b/shop/src/utils/generateVariants.js
@@ -32,7 +32,7 @@ export const generateVariants = (product) => {
   const existingVariants = (variants || []).reduce((obj, variant) => {
     return {
       ...obj,
-      // Using an unconventional separator as a form 
+      // Using an unconventional separator as a form
       // of serialization to keep things simple.
       [variant.options.join('|||')]: variant
     }
--- a/shop/src/utils/generateVariants.js
+++ b/shop/src/utils/generateVariants.js
@@ -32,6 +32,8 @@ export const generateVariants = (product) => {
   const existingVariants = (variants || []).reduce((obj, variant) => {
     return {
       ...obj,
+      // Using an unconventional separator as a form 
+      // of serialization to keep things simple.
       [variant.options.join('|||')]: variant
     }
   }, {})
--- a/shop/src/utils/generateVariants.js
+++ b/shop/src/utils/generateVariants.js
@@ -40,6 +40,10 @@ export const generateVariants = (product) => {
     (optionCombo, index) => {
       const comboTitle = `${product.title} - ${optionCombo.join(' / ')}`
 
+      const prevValues = {
+        ...existingVariants[optionCombo.join('|||')]
+      }
+
       return {
         title: comboTitle,
         name: comboTitle,
@@ -48,7 +52,7 @@ export const generateVariants = (product) => {
         available: true,
 
         // Default to previous values if it exists
-        ...existingVariants[optionCombo.join('|||')],
+        ...prevValues,
 
         // Set {option1, option2, ...} values
         ...optionCombo.reduce(
@@ -59,8 +63,8 @@ export const generateVariants = (product) => {
           {}
         ),
 
-        //Reset variant ID
-        id: index,
+        // Reset variant ID
+        id: product.externalId ? prevValues.id : index,
 
         // Set options array
         options: optionCombo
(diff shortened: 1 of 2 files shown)


shop/src/utils/generateVariants.js
```diff
@@ -40,6 +40,10 @@ export const generateVariants = (product) => {
     (optionCombo, index) => {
       const comboTitle = `${product.title} - ${optionCombo.join(' / ')}`
 
+      const prevValues = {
+        ...existingVariants[optionCombo.join('|||')]
```
franckc, line 44: I hadn't notice we use this interesting separator before :)
Maybe add a comment about why we need such a atypical separator?
Also where is this used? I did not see any reference to a split by '|||' anywhere in the dshop code?

shop/src/utils/generateVariants.js
```diff
@@ -59,8 +63,8 @@ export const generateVariants = (product) => {
           {}
         ),
 
-        //Reset variant ID
-        id: index,
+        // Reset variant ID
+        id: product.externalId ? prevValues.id : index,
```
franckc, line 67: Good catch!

franckc (approved):
Thanks for investigating and fixing this!
A few super minor comments inline otherwise LGTM

shop/src/utils/generateVariants.js
shahthepro: I just used that as a form of serialization, we didn't want to lose old values when a variant is added/deleted. Didn't want to complicate things, so went with a simple string concat. 

It's set and used in the same file: https://github.com/oplabs/dshop/blob/shahthepro/fix-883/shop/src/utils/generateVariants.js#L35

Will add a comment.

shahthepro:
@franckc Merging this now. Will run the script later today on prod to fix that shop's files. Let me know if you have any other comments

PR #896, opened 2021-02-04
Fix one-off script

shahthepro:
The script was failing on prod. After nuking my local `node_modules` and reinstalling deps, I was able to reproduce it. Guess it has to do with updated `commander` version in this PR #869. 

The other scripts might be broken too, should probably update them before we use any.


shahthepro:
Merging and deploying this now, will address any comments later.

franckc (commented):
Ah good to know. Yeah, we'll need to update the other scripts when we need to use them... Thanks for the heads up!

LGTM

Issue #912, opened 2021-02-14
Purchase not showing up on "Orders" page


gullible1:
A purchase was made on the shop https://graphics-illustrations-crypto-shop.ogn.app/ but the orders page still shows "no orders yet".

For reference, the tx is: https://etherscan.io/tx/0xf944dfae1fc83e97a46939c0900285091cabcb93766ad40a11c1938de7e9930a

franckc:
Re-ran the failed job and the order is now showing up in the graphics-illustration store.

Added more resiliency to our back-end in #913 so that this does not occur again.

PR #913, opened 2021-02-16
Add retries to the tx queue

franckc:
Fix for #912 

The tx queue was not configured to be retry-able. If a tx hash failed lookup because it was too fresh to show up on Alchemy, the order would fail.

Added retries to the tx queue and also cleaned up the makeOffer queue so that retries are configure at init time rather than having to do it when posting every job.

Bull's reference doc: <https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md>

### DIFF
--- a/backend/queues/queues.js
+++ b/backend/queues/queues.js
@@ -21,13 +21,18 @@ if (REDIS_URL) {
 }
 
 const all = [
-  new Queue(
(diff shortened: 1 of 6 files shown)


mikeshultz (approved):
LGTM

PR #924, opened 2021-02-26
Devops materials for new cluster

mikeshultz:
Adds new helm charts, dockerfiles, etc for new cluster.  Includes some improvements as well.

- Issuer (lua-resty-autossl) is now redundant and scalable.  It will use Redis to store certificates
- Deployments are simplified with [`helmfile`](https://github.com/roboll/helmfile) (generally won't be manual, but run via cloudbuild)
- Each network "node" is a helmfile environment, which are individual helm releases, both on the `default` namespace
- dshop instances are now Kubernetes Deployments and not StatefulSets, making scaling simpler/cleaner
- [helm secrets](https://github.com/OriginProtocol/dshop/pull/new/devnewops) now used for secret handling, which uses GCP keystore for encryption/decryption behind the scenes.  Simplifies editing and interaction (e.g. `helm secrets edit path/to/file.yaml`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,10 +31,10 @@ node_modules/
 
 # Deployment #
 ##############
-secrets.yaml
-secrets-dev.yaml
-secrets-staging.yaml
-secrets-prod.yaml
+#secrets.yaml
+#secrets-dev.yaml
+#secrets-staging.yaml
+#secrets-prod.yaml
 secrets-*.env
 github.key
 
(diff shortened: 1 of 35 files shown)


franckc (approved):
Eyeballed it and LGTM

When are we planning on doing the cluster switch? Let's coordinate. I'd prefer if we do that early next week as to not distract the team from the 3LAU end of week timeline in case an unexpected dshop issue shows up?

PR #925, opened 2021-02-26
Adds the ability to disable queue processing

mikeshultz:
Don't want a second set of queue processors running on separate state storage when I bring up the second mainnet environment.  This adds a way to disable them.


franckc (approved):
LGTM

Issue #653, opened 2020-10-20
Option to include/exclude taxes from discount 


gullible1:
Similar to how we allow users to choose whether to include/exclude shipping price from discounts, we should also let them do the same for taxes. 

For example: 

Currently if a seller sets a 100% off discount code for a Printful product, the buyer still has to pay the tax amount. The seller should have the option to cover taxes as well and have the buyer pay $0.

theprivileges:
@gullible1 could you help me understand this feature request a bit, please?  Is the desired outcome of this issue to:

- Add a checkbox to the _Discount_ page in `admin` where a shop admin can toggle _Exclude Taxes_.
- When the checkbox is selected, then no taxes are added to the cart total.
- When the checkbox is not selected, then taxes are calculated as expected, and added to the cart total.

Is this correct, or have I misunderstood your example?

franckc:
@theprivileges When the checkbox is selected, the discount should not apply to the taxes. But the taxes should remain in the cart. In some sense it is very similar to the existing option "Exclude Shipping Fees" but for taxes vs shipping.

For example, assuming both the "Exclude taxes" and "Exclude Shipping Fees" are selected. Here is an example of a cart:
Subtotal: $100
Taxes: $8.5
Shipping: $5
Discount 10% off: -$10
=================
Total: $103.5
Notice how the discount only applied to the subtotal and not to the taxes nor the shipping.

Let us know any question! 👍

phyninja:
Hi everyone,

I've implemented the requested feature in separate branch, and am in the process of writing a unit test to confirm that the cart totals are calculated as expected. I do have a concern to bring up at this point: 

If a seller chooses to configure Taxes on their UI (from Settings > Checkout), it seems to me like there isn't enough information for them to know how taxes are calculated. For example, how many new sellers would know that in order to generate @franckc's example cart, they would have to set the Tax Rate to 8.5%? In other words, are there places in the world where the norm is that a discount is applied on a product before calculating the tax on that 'final' product price?

franckc:
@micahalcorn @gullible1 Could you comment on ^^^ ?

phyninja:
@franckc since I haven't yet seen a response on this thread, I've enabled tooltips on the two checkboxes to inform shop admins of their functionality (screeenshots below). To view the tooltip, one just has to hover over their mouse over the option. 

Does this work for everyone? 

-----------
[...]

PR #939, opened 2021-03-19
Discount on taxes

phyninja:
#653 

Major change:
- Adds 'Exclude Taxes' option for Discounts.

Other changes:
- Adds tooltips for 'Exclude Shipping' and 'Exclude Taxes' checkboxes.
- Renames custom React elements to match their names in source files.
- Makes minor updates to documentation.


packages/utils/calculateCartTotal.js
```diff
@@ -68,7 +79,6 @@ const calculateCartTotal = (cart) => {
 
   const donation = get(cart, 'donation', 0)
 
-  const totalTaxes = Math.ceil(taxRate * subTotal)
```
franckc, line 71: Why did this line get removed but left in the return statement on line 93 below?
Have you checked there is no caller of this method relying on the totalTaxes field?

franckc (commented):
Took a quick pass at reviewing. One question inline...

packages/utils/calculateCartTotal.js
phyninja: The 'totalTaxes' field is still calculated -- see line 34 in the modified file. I made 2 changes in relation to it:
1) Moved it 'upwards' (to line 34) to aid in the calculation of the `prediscountTotal`.
2) Adjusted the math because `taxRate` is a "fixed point" number as mentioned in lines 29-32 of the modified file.

franckc: Ah... yep. My bad!

franckc (approved):
LGTM!

PR #952, opened 2021-04-14
fix new product selected collections

### LINKED ISSUE #930: Adding a product on the backend, collection selected lost
When adding a product and selecting the collection and saving the product, the collection selected is lostx.

baaae:
Fixes #930

When a new product was saved the selected collections would clear. This was caused by listening on changes to `collections.length`. After the product was pushed to the DB and fetched, the useEffect() would not update bc even though collections had changed `collections.length` did not change.


franckc (approved):
Thanks for fixing!

LGTM

Issue #933, opened 2021-03-05
Mobile checkout error


mikeshultz:
Looks like there's an error when trying to step through the checkout process.  This was noticed on https://store.originprotocol.com/.

On mobile, if you fill out your name and number and tap "Continue", nothing happens.  I haven't been able to look for console errors yet.

This was originally reported on Safari Mobile, and I was able to repro on Firefox Mobile (android).

etnlbck:
I was also able to repro Chrome iOS.

etnlbck:
@mikeshultz The bug has to do with how `form` validations are handled. 

On mobile, `Continue` should take the user to step 2, but appears to stick.
Since all of `state` gets validated when `onSubmit` is fired, the shipping fields aren't on the DOM causing the form to be invalid.

**Proposed Solution:**
If a user is on mobile, validate the `email` field then display shipping info fields. 
`shop/src/pages/checkout/Information.js`

etnlbck:
This issue can be closed.

PR #955, opened 2021-04-22
[#933] Mobile Checkout Bug

etnlbck:
The issue was that we are validating the whole state, instead of just email while on mobile. The fix is to pass which field to validate.


shop/src/themes/shared/checkout/Information.js
```diff
@@ -134,8 +134,7 @@ export const MobileInformation = () => {
 
   const onSubmit = (e) => {
     e.preventDefault()
-
-    const { valid, newState } = validate(state)
+    const { valid, newState } = validate(state, ['email'])
```
phyninja, line 137: Hi @etnlbck,

I worked this line into the [function definition](https://github.com/OriginProtocol/dshop/blob/master/shop/src/data/validations/checkoutInfo.js) for the `validate` function. Here's what I found:
Let's say the user (i.e. the "customer") is on a mobile phone and lands on the checkout page in consideration. They input nothing in the 'Email' field and hit 'Continue'. When they do so, the system calls the function `validate(state, ['email'])`, with
<pre>state.email = undefined</pre>
Stepping through `validate`, we first find
<pre>
> newState = {}
> newState = { emailError: some_value, firstNameError: some_other_value, ....., zipError: yet_another_value}
> filteredFields = ['emailError', 'firstNameError', .... , 'zipError']
</pre>

Evaluating the`filter` function in line 78 results in :
<pre>
filteredFields = ['']
</pre>

Because of this, `valid` will be set to `true` and the system will allow the user to continue to the next form.

Conclusion: One way to address this is to update the line to say `validate(state, ['emailError']`.

etnlbck: @phyninja great catch. I'll update.

phyninja (approved):

phyninja:
Given my knowledge of Origin's core team's current priorities, I surmise that @nick may not currently be in a position to add a review. This PR fixes an error that has been reported by multiple merchants; having reviewed it with the author, I feel confident that there will be no breaking changes to the production version of DShop. I'm therefore merging it to the 'master' branch.

PR #956, opened 2021-04-22
Check 'out of stock' items in checkout

nick:


franckc (approved):
Changes LGTM

Though I think there is still a possibility that a buyer could order an out of stock item, especially if there are a lot of purchases happening during a small time window. Because the back-end does not update the stock count synchronously during the purchase flow. Rather, it happens asynchronously either when we get the Stripe callback or when the crypto payment is confirmed on-chain.
See https://github.com/OriginProtocol/dshop/blob/8dac8eba04a81b4652363d6c23ddec2b0459d8a9/backend/logic/order/index.js#L531

PR #959, opened 2021-04-30
Component Memory Leaks

slightly-cooked:
This resolves react warnings about setting component state after the component has unmounted. 

Most of the memory leaks happened inside of hooks that had async operations. Once the async operation completed it failed to check wether the component was still mounted. It introduces a new hook `useIsMounted` that makes it reusable to check wether a component is mounted/unmounted.

Screenshot 2021-04-14 at 10 29 31 AM

### DIFF
--- a/shop/src/components/ConfirmationModal.js
+++ b/shop/src/components/ConfirmationModal.js
@@ -95,6 +95,7 @@ const AdminConfirmationModal = ({
             if (onClose) {
               onClose()
             }
+            setState({}, true)
           }}
         >
           <form
--- a/shop/src/components/ConfirmationModal.js
+++ b/shop/src/components/ConfirmationModal.js
@@ -95,7 +95,6 @@ const AdminConfirmationModal = ({
             if (onClose) {
               onClose()
             }
-            setState({}, true)
           }}
         >
           <form
--- a/shop/src/utils/useIsMounted.js
+++ b/shop/src/utils/useIsMounted.js
@@ -3,7 +3,7 @@ import { useEffect, useRef } from 'react'
 export default function useIsMounted() {
   const isMounted = useRef(true)
 
-  useEffect(() => () => isMounted.current = false, [])
+  useEffect(() => () => (isMounted.current = false), [])
 
   return isMounted
 }
--- a/shop/src/utils/useIsMounted.js
+++ b/shop/src/utils/useIsMounted.js
@@ -1,0 +1,9 @@
+import { useEffect, useRef } from 'react'
+
+export default function useIsMounted() {
+  const isMounted = useRef(true)
+
+  useEffect(() => () => isMounted.current = false, [])
+
+  return isMounted
+}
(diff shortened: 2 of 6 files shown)


shop/src/components/ConfirmationModal.js
```diff
@@ -95,7 +95,6 @@ const AdminConfirmationModal = ({
             if (onClose) {
               onClose()
             }
-            setState({}, true)
```
phyninja, line 98: Hi @baaae,

I reviewing this PR, I learned that this line functions as an `else` statement for the function definition between lines 91 and 98; I recommend keeping it because all components that use the `AdminConfirmationModal` component need to be updated with an `onClose` prop of their own otherwise. Examples include things like the [Delete button](https://github.com/OriginProtocol/dshop/blob/master/shop/src/pages/admin/discounts/_Delete.js) under **Discounts** and the [Publish Changes button](https://github.com/OriginProtocol/dshop/blob/master/shop/src/pages/admin/settings/deployments/_DeployButton.js).

The effect of not updating such components may be seen below:
AdminConfirmationModal

baaae: @phyninja Good catch! I'll add it back in for now and work out a different solution in the future.

shop/src/utils/useIsMounted.js
```diff
@@ -0,0 +1,9 @@
+import { useEffect, useRef } from 'react'
+
+export default function useIsMounted() {
+  const isMounted = useRef(true)
+
+  useEffect(() => () => isMounted.current = false, [])
```
phyninja: Hi @baaae,

I don't yet have a firm grasp on this code pattern. Could you please help me understand it by describing it via code comments, for example?

baaae: Sure. There's two thing going on here. 

One, we need to know when the component has unmounted. We do that by setting a cleanup function in the `useEffect`. We make sure the cleanup function fires only once, on unmount, by setting an empty dependency array.

Two, we need a way to pass around the value indicating wether the component is mounted or unmounted. With class components we could assign that value to an instance variable and use it throughout the lifecycle of the component. We're not using class components so even though that's a more elegant solution we can't use it here. Instead we use a [ref and treat it like an instance variable](https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables). 

Overall, when the component first mounts it sets the ref `isMounted.current = true` and when it unmounts it sets `isMounted.current = false`.

Hope that clears things up. Let me know if something is still fuzzy

phyninja (approved):

phyninja:
@franckc @nick 
Are either of you available to give this another look?

phyninja:
@franckc,
The author of this PR has clearly explained the reasoning behind the changes they made. Also, in my opinion, [there is enough support](https://stackoverflow.com/a/65152534) for the code pattern found in the attached commits to warrant their merge into DShop's source.
If you do not disagree, could you please merge the commits?

franckc (approved):
LGTM

PR #961, opened 2021-05-04
Case Insensitive Login

### LINKED ISSUE #928: Case Insensitive Login
Apparently, the login form is currently case-sensitive when it comes to email address. We should probably store them in lowercase, let the user type any case they want, and then convert to lowercase when querying the db.

baaae:
User-provided email is used in the following endpoints to either create a seller, update a seller, or compare to an existing seller. I checked all of these places to make sure emails were being converted to lowercase.

* user login ( `/auth/login` )
* superuser login ( `/superuser/login` )
* first time sign up ( `/auth/registration` )
* regular sign up ( `/register` )
* edit user ( `/superuser/users/{userId}` )
* forgot password ( `/forgot-password` )

Resolves #928

### DIFF
--- a/backend/db/migrations/20210504183956-update-email-lowercase.js
+++ b/backend/db/migrations/20210504183956-update-email-lowercase.js
@@ -1,0 +1,10 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface) => {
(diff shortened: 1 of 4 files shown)


franckc (approved):
LGTM

PR #962, opened 2021-05-05
Adjusting the formula that calculates total taxes

phyninja:
A user (Shop Admin) on Discord reported that "taxes are showing up as being thousands of dollars", when they "have it set at 6.25%". After confirming the issue on my dev environment (by setting a tax rate of 10% for a randomly chosen country), I'm proposing a solution for it via this PR.

#### Screenshot (before):
Tax-calc-before

#### Screenshot (after):
Tax-calc-after

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,3 +92,6 @@ devops/github
 *.dec.yaml
 
 dshop_setup
+
+# Other #
+.prettierrc.json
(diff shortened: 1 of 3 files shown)


phyninja:
This code was incorporated into the `master` branch via #939.

PR #966, opened 2021-05-13
Changes error message for a bit of clarity

mikeshultz:
Help narrow down the errors we're seeing a bit.


franckc (approved):
Looks DShopy to me!

PR #967, opened 2021-05-14
Incorrect hostname used in deployments

mikeshultz:
Looks like a not-insignificant chunk of shops are running into a domain conflict causing deploy failures after the recent addition of a validation check.  This uses `shop.hostname` instead of `dataDir` (auth_token) for deployment subdomain in the deploy route to help normalize things a bit more.  

This might not be the only cause of the conflict.  We were playing kinda fast and loose with DNS names.  Will continue to monitor and work on fixing.


franckc (approved):
LGTM

PR #969, opened 2021-05-17
Fix bad output for /resend-email

mikeshultz:
`/resend-email` should be sending a bool for `success`


franckc (approved):
Ah yeah. that explains why Sam reported this... thanks for fixing. LGTM

PR #972, opened 2021-05-21
Fix: Creating New Product

### LINKED ISSUE #971: No-op When Creating a Product
When trying to add a new product, clicking "Save" will fade the btn as if it's doing something but it stays stuck forever.

This is what the console shows:

Screenshot 2021-05-21 at 10 17 59 AM

Line 332 of shop/src/pages/admin/products/Edit.js calls `fetchProductStock.cache.clear()` when creating a new product, however `fetchProductStock` is no longer a memoized function after [this commit](https://github.com/OriginProtocol/dshop/commit/2122531de64ab59c7f5696086df16981b45130b2#diff-2f16a5ee4814f6f50fb7a90aabb506edf7bbfa5c0f5a476e0a8232e647e53b7d) so it no longer has the `cache` property.

baaae:
See #971 for details.

Closes #971

### DIFF
--- a/shop/src/pages/admin/products/Edit.js
+++ b/shop/src/pages/admin/products/Edit.js
@@ -329,7 +329,6 @@ const EditProduct = () => {
 
       // Clear memoize cache for existing product
       fetchProduct.cache.delete(`${config.dataSrc}-${product.id}`)
-      fetchProductStock.cache.clear()
 
       dispatch({
         type: 'toast',


phyninja (approved):
Makes sense.

phyninja:
@franckc @nick 
This PR needs a once-over as well.

Gratefully,
Rajath

phyninja:
@franckc @nick
I think this PR is worth a merge to `master`. Because of #971, a shop admin may not realize that the product they're trying to  "Save" is already so. Clicking on the _Save_ button more than once results in product duplication:

https://user-images.githubusercontent.com/10854442/154244253-4f8fd9c6-c4b1-4c51-92c4-e1363a4aecb5.mov

With the changes made in the attached commits, the DShop UX is improved:

https://user-images.githubusercontent.com/10854442/154245039-d6693d1f-6964-4d73-9344-4570811addac.mov

shop/src/pages/admin/products/Edit.js
```diff
@@ -329,7 +328,6 @@ const EditProduct = () => {
 
       // Clear memoize cache for existing product
       fetchProduct.cache.delete(`${config.dataSrc}-${product.id}`)
-      fetchProductStock.cache.clear()
```
franckc, line 332: Wasn't there a good reason for clearing the in-memory cache?
What is the impact of not clearing the cache anymore>

phyninja: > Wasn't there a good reason for clearing the in-memory cache?

`fetchProductStock` was a memoized function until it was determined that it being just an async function worked better for DShop: https://github.com/OriginProtocol/dshop/pull/956/commits/60255ad22383aaf2bda406565f84bc7481e3b3f7
My understanding is that a regular async function will not have a _cache_ property to clear.

franckc: ah yes, thanks for digging the details.

PR #978, opened 2021-07-30
Gate the logic for new shop creation using an env var

franckc:


mikeshultz (approved):
LGTM

Do we want to do any UI for this?  Or will it display the message just the same?

franckc:
> LGTM
> 
> Do we want to do any UI for this? Or will it display the message just the same?

I checked and it displays the error reason in the UI. So I'm hoping that's good enough.

Issue #958, opened 2021-04-26
Missing important pages


kevinp1:
Sales of goods require various policies to be available.  Individually, these include but are not limited to:

- Privacy Policy 
- T & C's
- Conditions of sale
- Returns policy

I suggest adding a new '_Policies_' option to the _Settings_ menu.

For each policy, shop owners should have the option to either, link to one or more of their existing policies using https URL's or, add new policies as editable text to be stored similar to _About_ page content.

Links to policies could then be accessed in the footer of themes.

An option can also be added in the _Settings > Checkout_ settings to toggle On or Off an 'I agree to terms and conditions'

phyninja:
Hi @kevinp1 (cc @franckc),

Because this idea of Dshop not supporting "mandatory" policy-related pages also being brought up [elsewhere](https://discord.com/channels/404673842007506945/809309667398385694/834303290065027094), I chose to take a second look at it. Given my familiarity with the 'CKEditor' module (the editor used by Shop Admins to edit their stores' **About** pages) however, I decided to pursue an alternative path towards a possible resolution here. 

The CKEditor module may allow us to expose a ['Source'](https://ckeditor.com/docs/ckeditor4/latest/features/sourcearea.html) button on the toolbar, allowing Shop Admins to edit the HTML Source Code of their shops' **About** pages. In the GIF below, I illustrate how one might leverage this feature to solve the main issue: 
![Mandatory pages](https://user-images.githubusercontent.com/10854442/116396491-18f64500-a843-11eb-9b44-0d993e6a1020.gif)

I'm checking in to establish consensus on one of the above ideas, because there are technical decisions to be made if this route is chosen. For example, security-minded folks may have concerns about Shop Admins embedding potentially-unsafe Javascript content in the **About** section of their shop.

phyninja:
Hi @kevinp1,

The editor found under **Settings** > **Appearence** already allows shop admins ("owners") to link to external webpages as shown in the screenshots below:
![1](https://user-images.githubusercontent.com/10854442/117824228-ee1cdf80-b28b-11eb-87eb-d3b5edbc2650.png)
![2](https://user-images.githubusercontent.com/10854442/117824279-fb39ce80-b28b-11eb-9162-14f34aaab5f2.png)

Does this suffice to meet the requirement of having "various policies [...] available"?

kevinp1:
Hi @phyninja 

[...]

PR #988, opened 2021-08-12
Feature - Allow shop admins to add 'policy' pages

phyninja:
Implements the feature requested in #958.

Note: This is currently available (only) on the 'default' theme.


backend/db/migrations/20210717101512-addPolicies.js
```diff
@@ -0,0 +1,35 @@
...
+        id: {
+          type: Sequelize.INTEGER,
+          autoIncrement: true,
+          primaryKey: true
+        },
+        shop_id: {
+          type: Sequelize.INTEGER
+        },
+        all_policies: {
+          type: Sequelize.JSON
```
franckc, line 16: nit: I think we favor using JSONB over JSON for postgres. though it is not support by sqlite. So in migrations we have been using the following pattern:
```type: isPostgres ? Sequelize.JSONB : Sequelize.JSON```
Let's use that as well here to be consistent?

File: backend/db/migrations/20210717101512-addPolicies.js [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async () => {
+      await queryInterface.createTable('policies', {
+        id: {
+          type: Sequelize.INTEGER,
+          autoIncrement: true,
+          primaryKey: true
+        },
+        shop_id: {
```
franckc, line 12: We should probably add an index on shop_id since this table will likely get queried by shop_id?

backend/routes/policies.js
```diff
@@ -0,0 +1,72 @@
...
+*/
+
+/*
+References:
+https://sequelize.org/master/manual/model-querying-basics.html
+https://expressjs.com/en/guide/routing.html
+*/
+
+module.exports = function (router) {
+  // Apply the callback 'authShop' to any POST request to the endpoint '/shop/policies'
```
franckc, line 17: nit: in express jargon, authShop is a `middleware`, not a `callback`

backend/routes/policies.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }).then((resultArray) => resultArray[0])
+
+    // To see the output of 'findOrCreate' in the Terminal:
+    // log.info(`policyEntry: `, policyEntry, `created? `, created)
+
+    // Sequelize's 'findOrCreate' method creates a new entry in the DB if one didn't exist. Verify that the data supplied by the frontend
+    // matches the policy data stored in the DB; update the DB if there is no match.
+    if (policyEntry.allPolicies !== req.body) {
+      res.send(
```
franckc, line 36: Let's sanitize the DB row object before sending it back on the network.
Something like:
```
 await polycyEntry.update({ allPolicies: req.body })
 return res.send(polycyEntry.get({ plain: true })
```

backend/routes/policies.js
```diff
@@ -0,0 +1,72 @@
...
+          },
+          {
+            where: {
+              authToken: req.shop.authToken
+            }
+          }
+        )
+      )
+    } else {
+      res.send(JSON.stringify('Nothing to post.'))
```
franckc, line 50: How about using: ```res.sendStatus(200)```

backend/db/migrations/20210717101512-addPolicies.js
```diff
@@ -0,0 +1,35 @@
...
+          type: Sequelize.INTEGER
+        },
+        all_policies: {
+          type: Sequelize.JSON
+        },
+        auth_token: {
```
franckc, line 18: I don't think we need the auth_token here. the shop_id should be sufficient.
The routes middleware will handle looking up the shop basd on the auth_token send by the front-end client.

backend/routes/policies.js
```diff
@@ -0,0 +1,72 @@
...
+    const data = await Policies.findOne({
+      where: {
+        authToken: req.shop.authToken
+      }
+    })
+    res.send(data.allPolicies)
+  })
+
+  router.delete('/shop/policies', authShop, async (req, res) => {
+    res.send(
```
franckc, line 64: In general, we don't want to send over the wire the result of DB operations (it may contain sensitive data). So I would call destroy, then after that call res.send(200)

franckc (changes requested):
Thanks for your contribution. A few comments inline regarding the DB schema and new APIs implementation. PTAL...

backend/routes/policies.js
phyninja: Updated.

backend/db/migrations/20210717101512-addPolicies.js
phyninja: If by "add[ing] an index on shop_id", you meant to suggest adding the line `primaryKey: true`, it is done.

backend/db/migrations/20210717101512-addPolicies.js
phyninja: Sources online recommend using JSON over JSONB when the (associated) information in the database is only subject to 'retrieval and storage' [[1](https://sequelize.org/master/manual/other-data-types.html#note-for-postgresql), [2](https://stackoverflow.com/questions/22654170/explanation-of-jsonb-introduced-by-postgresql)]. Because that is true of this case, could we please forego code consistency for the sake of efficiency in storage?

backend/db/migrations/20210717101512-addPolicies.js
franckc: You can't have 2 primary keys on a table. id is already your primary key. So you have to add an index on shop_id. something like `await queryInterface.addIndex('policies', ['shop_id'], { unique: true })`
Look into the dshop/backend/db/migrations directory, there are plenty of examples

backend/db/migrations/20210717101512-addPolicies.js
franckc: that's fine.

franckc (commented):
Took another pass. Thanks for addressing my feedback.

Could you please add some unit test for the new APIs added to the back-end?

phyninja:
> Could you please add some unit test for the new APIs added to the back-end?

Yes.

phyninja:
> > Could you please add some unit test for the new APIs added to the back-end?
> 
> Yes.

@franckc In my attempts of testing the routing code, I have so far been unable to diagnose and fix an issue with my unit tests. 
Here are the steps I have taken:
1) Create a test shop using the [createTestShop](https://github.com/OriginProtocol/dshop/blob/c40f4b05d7c0f1a9b6549b744d8080a964a92ada/backend/test/utils.js#L205-L257) utility
2) Verify that a shop is successfully created by logging the result
Excerpt:
```
Shop {
  dataValues: {
    enableCdn: false,
    id: 1,
    name: 'TestShop1630411318481',
    networkId: 999,
    listingId: '999-001-0',
    authToken: 'shopAuthToken1630411318481',
    config: '3a04e8418991fc91b7148c79a9618b75:4faa1d69a7c8225e8b03da0db8dce649ab2b9f5f89d78e32dc68e29cdecbc6ff84578e330f58872472dab94edfe359e7d4f3f3efd74393740fef1a0518b5f8c4dc7bf10ee0430b79b8b9f5333d7e298187d1f10342406d7dccc841b513585ec70c554754edd4536d8ae44cb9933bd61b5d8d4be61d323989a521c7eecfbe51620e5a007b254df15402eec5e8f5c98c71b3546eb0b26944343e1f97e24460da106a3d96b87f258f76eee4030803d6202dbf0fe77b363394b5ea5391ff1c4c961e6ecf7a7c5719f9c9a798ab89851cd26ec9ac4ba6127d78dd841f3a31d2bfe068af32b7b989b8873f226d935b0c6267c7be57f4533767d54d2e8a2b435215bda2842bd37640aba9fecd0f76f66007c5fe580b61f96efc0929035997ae7f0b82eb0a3e00b4aecba3b807454b85bcfe41e98eeedaa5a2addcd21028b6bafa71adcf7f4b2c48f26daf0355b437843db7a37c8a1fd5ba164769380efc3c9e02303dda951176d7fdbfd465dc36803b77dfbf31d53a47037d1e0c63207a789237a0020ceba4de0f5086f0b244bd094ab09b8d03cf5468fb1bc1e2fb080d6ce01de49585e3238b58888b56413cd0e08e21cb2fc132d2cffc71452a9c4408e07ba95c7bbd626ad69dc5c9212261d0e53fa460ed4d54c7ac38baac51b23365ab1cef951fdafd069f16dac5fb9fa8305d3bfe5f02458722077bcf94da9e55758d0c41a34cec9b9d6ac9e35e0c24b572dd38fbcc64003e0b26e7408cf847bf504542f89ca49a81784d8163911d928ad6ecd3f241b82849617b834e34750a2dc65a0456f68d5f8dc01948e698ecfb0e53fa47c132e1fbd6acf18a59d15cd00878e91994ff686e02c20357af1045d09aa2c3008ef296a3f726f6be43272399594fc29644f34e9e68bc4fe0b728f7891007e31d3a86101d5e1a482fc70dbfea52661d1a3ae53a397bc74185204ae03552538b28caf1ab660923e5e7cbceaa11ad7372ce029fcee2b87b9a0137b30c51206b6dbf1383d173f5754aa8bb883d45d049cc74f505ed6b599bc3a08c66e040e77af642272c9c73dac0ea38967c47a046de19299160dca2815e7dde400706eb548680830409b44d3e9ef1fda410ca9fb0c75a739605f867073f7fa2190fbade43ad3c4b0ed9236b1bd917c501f7b3c3d86bfb26e2611c5fe8600ac64b094b5d3a10f354e53c246daaf7365813f855f1702ee458a8e835a98dc1f468a0d59fea72aec33d871fe57d11785980f08eee1a9d878f634214803c23fe45b25f1e46fdd811641fe75a7f7268f5764d58c8e9bee69b4a654b3a8f0bbd20effd9e98463db6333a380e97a95e64f1b3e763bd5bb2284849df19a3cf41ec9f429a9e02eace6e35a1dbb0a8b6d95220f0b7ea14885d1ba807b3cdee8ef147db21508e44540bf09f9b38db8c3ca2cd7273cb56b05a1ff6086e2a7444f06def2423ab9ee91fa7998d2447eda4b09db281a3c7d52f60778b7b17ea408633b61dac20d9469c8ce9649ed3d8e6bd3f09498d3d53c678fd6602366dc91ef351a94a7919062f3d5ef2962ead2e2df5b220956acb6fff9724d76ca6b02c9e322bdbd6478ee2a1d9f5ef8519459e634c0009b2cc5f3ba38f83df8564718c64c6c839c0867d4c72ec1f9e6a39a38a492a1d85f568defd41f1a0540c2ee4cfb248a4d6b34fd51c6e833ecc9da8c7ca5612d8cf08b70879d790ac7ead367445eae4771f01924e7a66a9cdfca97c3db1235e67092077282cf896d1f0f9dfe70a67fb1b53079bc10069c43ea8fa7a9d8f470cceb82007290805946fbbb8513ee50340dbe5e6e8a637712d573a85e0239ccb2be49660e4d168572539add35c5b9314cb1c7b569b89d0fd2cff2dfb6af42d1499b5d4d5a71b44891f299e68657caa79b29aed87403a509bdbfe9cab6a64f50d11ce30695dc6dc58f6726121152660238dbaf493f10fa41611493c2f8869c1997797728643da8a352279a2a24360c8ef5eacf8159e1769b5654dc62f00f2188bbd80991c3a41336fe1d653f90e08a630280acdd3526efd12fc93d0bc8108163777556d40531e57bb29f4c05ecacae2a37ea2161ef18407201c986ce11b060defa77b9117cd3bf80b34785d52edd1c566b75e438e0fdb167cb012c0a221eba0bfe1df2a6ce51b23166cf26840ed7c4d8283bb98d15074b920584dcde4bc0aca5c7facb951ef648a2ee3ca2af856a4ae3e7e15cb710f66467eadccfd65bbea64ee005e96afc7ddd33470e0efb3755fcb56261a7e45493a139423a66456817d8a288a7140dc0f8f62af1bcb3b84a42471d2c46d9dd63a2b7645e9ae74d080fd96d01dc2e331f5e0a57d75b1ac1b9929bf7b3f1f79205b38fdb236c4680def5ba0e727a814de0253d97897183094ae037cdc8996fe79f675cf745cbdbb860681457fea461af8f9b794098a158a16101d0fc0f2e0765252839c9088cf7b6109ec244c234517',
    sellerId: 1,
    hostname: undefined,
    updatedAt: 2021-08-31T12:01:58.482Z,
    createdAt: 2021-08-31T12:01:58.482Z
  }

```
3) Test the 'post' method using
```
const fetch = require('node-fetch')
const { ROOT_BACKEND_URL } = require('./const')
const testPolicies1 = [['', '', false]]

const query1 = await fetch(`${ROOT_BACKEND_URL}/store/policies`, {
	method: 'post',
	headers: {
		authorization: `bearer ${encodeURIComponent(
			shop.authToken
		)}`,
		'Content-Type': 'application/json'
	},
	body: JSON.stringify(testPolicies1)
})

```
In the (resulting) response, the following fields may be noteworthy:
```
url: 'http://localhost:8357/store/policies',
status: 404,
statusText: 'Not Found'
```
I'm confident that the app is running at `http://localhost:8357` during the tests, despite being unsure of why GET and POST requests to the 'newly created' endpoints are not working.

phyninja:
@franckc 
It turned out that I was making GET and POST requests to a non-existent endpoint ('/store/policies') when I should've been testing '/shop/policies'. Please let me know if there's anything else to do.

phyninja:
@franckc are you available to take another look?

phyninja:
@franckc 
Considering that the feature attached to this PR may provide utility to DShop admins like @kevinp1, could you please provide [us] an update in case you're unable to review this?

shop/src/pages/admin/settings/legal/Edit.js
```diff
@@ -0,0 +1,321 @@
...
+
+        try {
+          // Make sure that the request body is either a JS Object or Array. Failure to do
+          // so will produce a SyntaxError when the body is parsed by Express' body-parser
+          // module (i.e. backend). Discussion: https://github.com/expressjs/body-parser/issues/309
+          const shopPoliciesRes = await post('/shop/policies', {
+            body: JSON.stringify(policies),
+            suppressError: true
+          })
+          shopPoliciesRes ? null : console.log(`Error posting shop's policies`)
```
franckc, line 182: Shouldn't an error get displayed in the UI in case the API call returns an error?

backend/routes/policies.js
```diff
@@ -0,0 +1,61 @@
...
+    // Step two: Using the result of the previous function call, find or create an instance of the model 'Policies' in the DB
+    // Sequelize's 'findOrCreate' method creates a new entry in the DB if one didn't exist.
+    const policyEntry = await Policies.findOrCreate({
+      where: {
+        shopId: req.shop.id
+      },
+      defaults: {
+        allPolicies: [['', '', false]]
+      }
+    }).then((resultArray) => resultArray[0])
```
franckc, line 28: Mixing await and .then makes the code hard to comprehend. Let's not mix await and .then

backend/routes/policies.js
```diff
@@ -0,0 +1,61 @@
...
+    // log.info(`policyEntry: `, policyEntry)
+
+    // Step three: Regardless of whether or not a new entry was created, confirm that the data supplied by the frontend matches the policy data stored in the DB by calling Sequelize's 'update' method.
+    if (policyEntry.allPolicies !== req.body) {
+      const updatedPolicyEntry = await policyEntry.update({
+        allPolicies: req.body
+      })
+      res.send(updatedPolicyEntry.get({ plain: true }))
+    } else {
+      res.sendStatus(200)
```
franckc, line 40: Why not returning the policy to make the API consistent?
Otherwise the caller would have to deal with either empty result or result with the policy.

backend/test/policies.test.js
```diff
@@ -0,0 +1,122 @@
...
+    ],
+    ['Return Policy', 'Eget egestas purus viverra accumsan.', false]
+  ]
+
+  const testPolicies3 = []
+
+  it('Should post new shop policies', async () => {
+    //If an admin just clicks on the 'Update' button without adding any policy, the server should just return a '200' response
+
+    await chai
```
franckc, line 54: Thanks for adding those tests!
Sorry for being picky, but lould we keep the codebase consistent and use the same patterns as in all the other tests in the backend/tests directory and not used chained calls like await foo().request().post()...?

shop/src/pages/Main.js
```diff
@@ -82,13 +104,39 @@ const Main = () => {
...
+          return res.json() //Expected type: <Array<Array<String, String, boolean>>>
+        }
+      })
+      .then((result) => {
+        if (result) {
+          setPolicies(result)
+        }
+      })
+      .catch((err) => {
+        console.error('Failed to load shop policies', err)
```
franckc, line 128: Should display an error in the UI

shop/src/pages/admin/settings/legal/Edit.js
```diff
@@ -0,0 +1,321 @@
...
+              </fbt>
+            )
+          })
+        } catch (err) {
+          console.error(err)
+          setSaving(false)
+        } finally {
+          // Notify the backend that there have been changes to the shop's config. This is done to ensure that the deploy (i.e. 'Publish Changes') button is displayed
+          // to the user
+          const shopConfigRes = await post('/shop/config', {
```
franckc, line 201: The implementation of the /shop/policies route should be modified to take care of that. The client should not have to make this API calls.

franckc (commented):
Took another pass. I still have some concerns. Comments inline.

backend/test/policies.test.js
phyninja: Yes.

shop/src/pages/admin/settings/legal/Edit.js
```diff
@@ -0,0 +1,321 @@
...
+      .then((res) => {
+        if (!res.ok) {
+          return null
+        } else {
+          return res.json() //Expected type: <Array<Array<String, String, boolean>>>
+        }
+      })
+      .then((result) => {
+        if (result) {
+          timeout = setTimeout(() => setPolicies(result), 2000)
```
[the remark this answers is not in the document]
phyninja: I'm not sure; the 'useEffect' hook here is adapted from another part of the codebase -- [one dealing with the CKeditor module under Settings/Appearence](https://github.com/OriginProtocol/dshop/blob/4835debe63f3f009a01b200eff59789cc2d822b8/shop/src/pages/admin/settings/appearance/Edit.js#L57-L73). Unless you insist, I will refrain from delving further into the reason for the timer's existence.

shop/src/pages/Main.js
phyninja: This fetch happens when a DShop customer lands on the main (public-facing) page of the shop. In the scenario where an error occurs during the fetch, I just expect that the footer (of the main page) display links to everything but the shops policies -- for example, **About** and **Contact**. If you still think an error should be displayed in the UI, could you please share your reasoning?

shop/src/pages/admin/settings/legal/Edit.js
franckc: Sorry un-resolving since I did not see a resolution for this?

And actually, thinking more about it, I have a more fundamental question. Why do we actually need to re-publish the shop given that the way you implemented getting the policy is by making a dynamic API call? Since the policy content is dynamically fetched by the client and served by the back-end DB, we shouldn't need to re-publish the store?

franckc (commented):
Still a few questions... See inline.

shop/src/pages/Main.js
franckc: Got it. Yes, that's reasonable. An error causing the footer to not display is probably not worthwhile surfacing to the user. Thanks for the explanation.

shop/src/pages/admin/settings/legal/Edit.js
```diff
@@ -0,0 +1,321 @@
+import React, { useState, useEffect } from 'react'
+import fbt from 'fbt'
+import { get /*pick, pickBy*/ } from 'lodash'
```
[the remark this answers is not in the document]
phyninja: I think you were viewing an 'outdated' version in the commit history. Let me mention your username elsewhere to show you an up-to-date file.

Ref: [commit 5215d41](https://github.com/OriginProtocol/dshop/pull/988/commits/5215d41da3e57fcb57e1fc93928155bc835ed4ae)

shop/src/pages/admin/settings/legal/Edit.js
```diff
@@ -1,9 +1,8 @@
 import React, { useState, useEffect } from 'react'
 import fbt from 'fbt'
-import { get /*pick, pickBy*/ } from 'lodash'
+import { get } from 'lodash'
```
phyninja, line 3: @franckc

shop/src/pages/admin/settings/legal/Edit.js
phyninja: > Sorry un-resolving since I did not see a resolution for this?

I made a couple of changes to move this logic to the API code ([commit 5215d41](https://github.com/OriginProtocol/dshop/pull/988/commits/5215d41da3e57fcb57e1fc93928155bc835ed4ae)), but after reading what you say afterwards, I see your point.

franckc (approved):
Thanks for the demo video. Looks good!

backend/package.json
```diff
@@ -38,6 +38,7 @@
     "bs58": "^4.0.1",
     "bull": "^3.20.0",
     "bull-board": "^0.10.0",
+    "chai-http": "^4.3.0",
```
franckc, line 41: This should be moved to dev-dependency since it is only used for testing.

franckc (changes requested):

franckc (approved):
LGTM!

Issue #990, opened 2021-08-14
OpenPGP error preventing checkout in Preview mode


phyninja:
#### Steps to reproduce
1) Select a theme [for example, "Galleria"]
2) Checkout a product
3) Select a cryptocurrency as payment
4) Click 'Pay'

**Expected behavior:** Order is created successfully.

**Actual behavior:** The 'Pay' button changes to 'Processing', but nothing else happens [screenshot attached].
![OpenPGP error](https://user-images.githubusercontent.com/10854442/129440944-2a85ad55-220e-4bc9-80e8-9a918df68a3f.png)

phyninja:
This issue is linked to the code block https://github.com/OriginProtocol/dshop/blob/4e4f257ce5d7ecb39b5e61717d21a2320346e6d2/shop/src/utils/usePGP.js#L13-L21
where the web browser is instructed to load the PGP script from the server, with a _relative_ path of `dist/openpgp.min.js`. While the script may load successfully for the 'customer', it will fail for an admin previewing their shop under a theme. 

#### Reason: 
Inspecting the console as shown in the screenshot below, it is possible to see that the full path of the GET request is not found on the server.
![OpenPGP error](https://user-images.githubusercontent.com/10854442/129721638-5348193c-7c33-4f19-b8fd-e659e9d13beb.png)

PR #993, opened 2021-08-24
Updated the "usePGP" module

phyninja:
Addresses #990 by using absolute paths while loading the OpenPGP script.


franckc (commented):
It may fix the checkout flow in preview mode but won't that break it in production?

CC @nick sicne he is more familiar with this code than I am...

phyninja:
@franckc 
In the latest commit, I have introduced a workaround wherein the required OpenPGP module files are copied into each theme's _dist_ folder at build time. Would you please review?

franckc:
@nick You are more familiar than I am with this area of the code. Does that look good to you?

nick (approved):

PR #1018, opened 2021-11-23
[UI Update] Checkout flow for DShop's themes

phyninja:
### Order Summary
Attached are images that show the difference(s) between the way the 'Order Summary' is displayed on themes. The left side of an image is from the code on the `master` branch.

#### Theme name: Art/"Minimal"
Comparison - Art: Minimal

#### Theme name: Poly/"Showcase"
Comparison - Poly: Showcase

### Shopping Cart
On the _Poly/"Showcase"_ theme, a customer may not be able to _see_ the icon to remove an item from their shopping cart. Furthermore, the _dropdown to update an item's quantity_ could be improved:
Shopping Cart Comparison - Poly: Showcase


shop/src/themes/shared/checkout/_Summary.js
```diff
@@ -87,19 +86,29 @@ export const OrderSummary = ({ cart, discount = true }) => {
...
+        <span className="absolute -top-3 -right-3 h-6 w-6 rounded-full bg-gray-600 text-sm text-white flex justify-around items-center dark:border-gray-700">
+          {quantity}
+        </span>
+      </div>
+      <div className="flex flex-col justify-center">
+        <div className="font-semibold">{title}</div>
+        <div className="text-xs font-light">{productOptions}</div>
+      </div>
+      <div className="flex items-center justify-end font-semi-bold dark:border-gray-700">
+        {formatPrice(quantity * price, currencyOptions)}
```
phyninja, line 110: Reason for changing the props passed to `Row`:

When a user checks out, they are first shown the page where they can modify their cart if need be (Example: https://user-images.githubusercontent.com/10854442/143019839-698416b0-07d2-452d-84d0-af5981f02f4f.png). The price displayed on each row here is calculated -- `quantity * (price per item)`.
On the next screen, when they're asked to input their contact and shipping details, we display an **Order Summary** on the right, but do not format the price in the same way as before. 

Reference: https://github.com/OriginProtocol/dshop/blob/f65491f003fa96271f9b6f7c6554d05ae039e737/shop/src/themes/shared/Cart.js#L45-L51

